PR #3, opened 2026-04-11
Fix: ci.yml and llvm-build.yml

fabianlim:
# Fix: cmake-test and llvm-build

## Test Status

| Part | Workflow | `mlir-source` | Speed | Status |
|---|---|---|---|---|
| 1 | `ci.yml` → `cmake-test` | `mlir_wheel` | ⚡ Fast — prebuilt wheel, no LLVM compile | ✅ Passed |
| 2 | `llvm-build.yml` → `build` → `trigger-ci` | `custom` | 🐢 Slow — full LLVM build (~1h, ~1GB artifact) | ✅ Build passed — see [github action](https://github.com/torch-spyre/ktir-mlir-frontend/actions/runs/24288875086/job/70922784508); `trigger-ci` → `cmake-test` cannot be tested until merged |
| 3 | `ci.yml` → `cmake-test` | `custom` | ⚡ Fast — reuses artifact from Path 2 | ✅ Passed — [fork workflow_dispatch](https://github.com/fabianlim/ktir-mlir-frontend/actions/runs/24349727570) (but still need to address artifact freshness issue) |

The two skipped checks (`build`, `trigger-ci`) will always appear in PRs — they only run when an LLVM build is needed.

---

## Path 1: `mlir-source=mlir_wheel` (tested)

Three fixes were needed, all caused by `uv` or `mlir_wheel` triggering an unintended cmake build:

- **`uv sync --no-install-project`** and **`uv run --no-project`**: without these flags, uv installs the project in editable mode via scikit-build-core before running any command. This triggers a cmake build with `-DKTIR_ENABLE_PYTHON_BINDINGS=ON`, which fails because python source files are absent on this branch.
- **`CMakeLists.txt` guard** (from `python-packaging` branch): `mlir_wheel`'s `MLIRConfig.cmake` unconditionally sets `MLIR_ENABLE_BINDINGS_PYTHON=ON`, causing nanobind detection to run even with `-DKTIR_ENABLE_PYTHON_BINDINGS=OFF`. Fixed by introducing a project-local `KTIR_ENABLE_PYTHON_BINDINGS` option as the guard instead.

---

## Path 2: `mlir-source=custom` — LLVM build (partially tested)

`llvm-build.yml` builds LLVM from a pinned commit and uploads a GitHub Actions artifact
(`llvm-<short-hash>-<os>-<arch>.tar.gz`). Two ways to trigger:

- **`push` on `cmake/llvm-hash.txt`**: fires automatically when the hash changes on `main`. Not active yet — file not committed. This is the flow Triton adopts.
- **`workflow_dispatch`**: provide the full 40-char LLVM SHA in the `llvm-hash` input.

Tested in this PR by: Actions → LLVM Build → Run Workflow → hash `3b3ac5a1169722bff1ae0f5f8f27a48cc08c3d02`
- Build ran and artifact was cached. Total size was **~1GB** — worth trimming in future (strip static libs, disable unused tools).

<details>
<summary>Artifact stored in torch-spyre/ktir-mlir-frontend</summary>

```json
{
  "id": 6388171867,
  "name": "llvm-3b3ac5a1-ubuntu-x64",
  "size_in_bytes": 1057486279,
  "archive_download_url": "https://api.github.com/repos/torch-spyre/ktir-mlir-frontend/actions/artifacts/6388171867/zip",
  "expired": false,
  "expires_at": "2026-07-10T18:32:28Z"
}
```
</details>

Build artifacts can be potentially distributed for local development:
- avoiding mindless rebuilding amongst users

```
curl -L -H "Authorization: token $GIT_PAT" \
    https://api.github.com/repos/torch-spyre/ktir-mlir-frontend/actions/artifacts/6388171867/zip \
    -o llvm-3b3ac5a1-ubuntu-x64.zip
```

---

## Path 3: `mlir-source=custom` with pre-built artifact (tested)

This path has `ci.yml` download the artifact produced by `llvm-build.yml` and run
Configure/Build/Test against it. Tested via `workflow_dispatch` from a fork:
[fabianlim/ktir-mlir-frontend run 24349727570](https://github.com/fabianlim/ktir-mlir-frontend/actions/runs/24349727570).

`Resolve build configuration` step queries the GitHub API to
auto-resolve the artifact repo (parent if fork, self otherwise) and look up the `run-id`
by artifact name. No extra inputs required. Works for `workflow_dispatch` from a fork,
`pull_request`, and `workflow_call` from `llvm-build.yml`.

### Problem: Artifact freshness

Retention is 3 days. If the artifact expires before `ci.yml` is triggered with
`mlir-source=custom`, the download step fails. Options:

- **Scheduled rebuild** (e.g. every 2 days): keeps artifact alive via `overwrite: true` + retention reset. sccache makes reruns cheap. Requires `cmake/llvm-hash.txt` on `main`; GitHub auto-disables schedules after 60 days of inactivity.
- **Increase retention** (e.g. 7–30 days): simple but higher storage cost.
- **Self-provisioned store** (S3, Artifactory, Azure Blob): no expiry, no quota concerns. Replaces `upload/download-artifact` with `curl`/`rclone` + credentials in Secrets. Triton uses Azure Blob for the same reason.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,10 +72,10 @@ jobs:
       - name: Set paths (mlir_wheel)
         if: ${{ inputs.mlir-source != 'custom' }}
         run: |
-          MLIR_DIR=$(uv run python -c \
+          MLIR_DIR=$(uv run --no-project python -c \
             "import mlir_wheel, pathlib; \
              print(pathlib.Path(mlir_wheel.__file__).parent / 'lib/cmake/mlir')")
-          LIT=$(uv run which lit)
+          LIT=$(uv run --no-project which lit)
           echo "MLIR_DIR=$MLIR_DIR" >> $GITHUB_ENV
           echo "LIT=$LIT"          >> $GITHUB_ENV
 
@@ -92,7 +92,7 @@ jobs:
         run: |
           tar -xzf llvm-${{ inputs.llvm-short-hash }}-${{ matrix.config.os }}-${{ matrix.config.arch }}.tar.gz
           MLIR_DIR="$(pwd)/llvm-${{ inputs.llvm-short-hash }}-${{ matrix.config.os }}-${{ matrix.config.arch }}/lib/cmake/mlir"
-          LIT=$(uv run which lit)
+          LIT=$(uv run --no-project which lit)
           echo "MLIR_DIR=$MLIR_DIR" >> $GITHUB_ENV
           echo "LIT=$LIT"          >> $GITHUB_ENV
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ on:
           - mlir_wheel
           - custom
       llvm-short-hash:
-        description: "Short LLVM commit hash (required when mlir-source=custom)"
+        description: "First 8 characters of the LLVM commit SHA (required when mlir-source=custom)"
         required: false
         type: string
   workflow_call:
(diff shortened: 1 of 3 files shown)


fabianlim:
GitHub-hosted runners are Ubuntu, Windows, and macOS only. No RHEL. Currently `llvm-build.yml` is setup for ubuntu.

  For RHEL-compatible builds, the common approach (which Triton uses) is to run an AlmaLinux container on an Ubuntu
  runner:

  runs-on: ubuntu-22.04
  container: almalinux:8.10-20250411

  AlmaLinux is a 1:1 RHEL binary-compatible rebuild, so it covers the RHEL use case without needing a licensed RHEL
  runner. For actual RHEL you'd need a self-hosted runner. For Ubi9 then AlmaLinux 9 is a closer match

ashokponkumar:
@HarikrishnanBalagopal ^

mudhakar (dismissed):
lgtm.. adding a issue ARTIFACT_REPO empty check

ashokponkumar:
>  GitHub-hosted runners are Ubuntu, Windows, and macOS only. No RHEL. Currently `llvm-build.yml` is setup for ubuntu.
> 
>   For RHEL-compatible builds, the common approach (which Triton uses) is to run an AlmaLinux container on an Ubuntu
>   runner:
> 
>   runs-on: ubuntu-22.04
>   container: almalinux:8.10-20250411
> 
>   AlmaLinux is a 1:1 RHEL binary-compatible rebuild, so it covers the RHEL use case without needing a licensed RHEL
>   runner. For actual RHEL you'd need a self-hosted runner. For Ubi9 then AlmaLinux 9 is a closer match
> 

@HarikrishnanBalagopal and Gnanesh are in the process of adding self hosted runners for image build, which can be used.

fabianlim:
@mudhakar thanks for the review. I addressed #5 in the above commit https://github.com/torch-spyre/ktir-mlir-frontend/commit/c082c4017e6562f2796f6b16cc1f93b14f9e900f

mudhakar:
@ashokponkumar need your approval.

ashokponkumar (approved):

Issue #12, opened 2026-04-22
[Feature] Publish ktir-mlir-frontend Python wheels


fabianlim:
cc: @mudhakar @raghukiran1224 @kiszk 

## Summary

The Python wheel build is fully wired in CI — `ci.yml` builds and installs the wheel on every PR (PR [#1](https://github.com/torch-spyre/ktir-mlir-frontend/pull/1), [#7](https://github.com/torch-spyre/ktir-mlir-frontend/pull/7)), and `llvm-build.yml` produces a wheel from the pinned LLVM hash in `cmake/llvm-hash.txt`. None of these wheels are published anywhere. This issue proposes adding a publish step so downstream repos (e.g. `ktir-cpu`) can install `ktir-mlir-frontend` as a plain Python package without any MLIR/LLVM build dependency.

## Why the wheels are self-contained

The ktir wheel bundles the compiled C extension (`_mlir_libs/*.so` / `*.dylib`) produced by scikit-build-core. MLIR is statically linked into that extension at build time. Consumers only need CPython — no `mlir_wheel`, no LLVM, no CMake. The wheel is ~19 MB compressed / ~70 MB uncompressed; `libKtdpPythonCAPI.so` dominates at ~34 MB due to statically linking all MLIR dialect C API libraries.

### Option A — GitHub Actions artifacts with a refresh workflow (recommended first step)

Two cooperating pieces:

**Part 1 — upload on every CI build (`ci.yml`)**

After `Build and install Python wheel`, build the wheel into `dist/` and upload it as a short-lived artifact (3-day retention, same as the LLVM artifacts). The artifact name encodes the LLVM hash (`ktir-mlir-frontend-<llvm-short-hash>-<os>-<arch>`) so consumers know exactly what LLVM commit the wheel was linked against. Before uploading, check via the GitHub API whether an artifact with that name already exists for this run; if so, skip the upload to avoid duplicate-artifact conflicts.

```yaml
- name: Build wheel
  run: CMAKE_ARGS="-DMLIR_DIR=${{ env.MLIR_DIR }}" uv build --wheel

- name: Check if wheel artifact already exists
  id: check-artifact
  env:
    GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  run: |
    ARTIFACT_NAME="ktir-mlir-frontend-${{ steps.config.outputs.llvm-short-hash }}-${{ matrix.config.os }}-${{ matrix.config.arch }}"
[...]

PR #1, opened 2026-04-11
[1/3] Python Wheel Packaging

fabianlim:
## What this PR does

This PR adds a complete Python packaging story to the project:

1. **A self-contained installable wheel** — `pip install` / `uv sync` builds and
   installs the `mlir_ktdp` Python bindings + all native libraries. No separate
   MLIR installation required at runtime.

2. **A pure-Python utilities package** (`tools_ktdp`) — ships alongside
   `mlir_ktdp` in the wheel. Fully live-editable: changes to Python files take
   effect immediately without rebuilding.

3. **A `ktdp-walk` CLI tool** — inspect any `.mlir` file's operation tree from
   the command line.

4. **A pytest suite** — three parametrized fixtures covering dialect
   registration, flat KTDP ops, and nested control flow.

---

## How it installs

The build backend is [scikit-build-core](https://scikit-build-core.readthedocs.io/),
which drives cmake under the hood. `uv` manages the virtual environment and
locks all dependencies.

```
uv sync --extra mlir
```

That single command:
- Creates/updates the venv
- Installs `mlir_wheel` (provides MLIR cmake config + Python support libs)
- Runs cmake → builds the C++ extension + CAPI library
- Installs the project in **editable** mode

---

## Frozen dependency versions (`uv.lock`)

All versions are pinned in `uv.lock` for reproducible installs:

| Package | Pinned version | Role |
|---|---|---|
| `mlir-wheel` | `20260409+3b3ac5a11` | MLIR cmake config + bundled Python support libs |
| `nanobind` | `2.12.0` | Python ↔ C++ binding library (runtime dep) |
| `pytest` | `9.0.3` | Test runner (`--extra test`) |
| `lit` | `18.1.8` | LIT test runner (`--extra test`) |

`mlir_wheel` is an **optional** dep — users with a custom MLIR build can pass
`CMAKE_ARGS="-DMLIR_DIR=..."` and skip it entirely.

---

## Wheel contents

The wheel is self-contained (~19 MB compressed, ~70 MB uncompressed). No
`mlir_wheel` needed at runtime.

```
ktir_mlir_frontend-0.1.0-cp312-cp312-linux_aarch64.whl
│
├── mlir_ktdp/                         ← cmake-managed
│   ├── __init__.py
│   ├── ir.py, passmanager.py, rewrite.py
│   ├── dialects/
│   │   ├── ktdp_nanobind.py           ← KTDP dialect Python bindings
│   │   ├── builtin.py, ext.py
[...]

PR #14, opened 2026-04-23
[Step 2/3] Pin LLVM 22.1.3: stable artifact CI and setup_mlir.py developer flow

### LINKED ISSUE #11: [CI] Add macOS ARM64 and Ubuntu ARM64 to `ci.yml` for Stable LLVM versions
@mudhakar @raghukiran1224 

## Summary

Add CI coverage for macOS ARM64 (`macos-15`) and Ubuntu ARM64 (`ubuntu-22.04-arm`) via a new `stable` job in `ci.yml`. The existing `cmake-py-test` job (Linux x64, nightly `mlir_wheel` and `custom` paths) is untouched.

The motivation is PR [#9](https://github.com/torch-spyre/ktir-mlir-frontend/pull/9) ("Prevent linker optimization affecting dialect registration"), which surfaced a macOS-specific build failure invisible to Linux CI: Apple's `ld64` discards symbols referenced only by static initializers, causing dialect registration to silently fail at runtime. The fix required platform-specific linker flags (`-Wl,-all_load` on macOS vs. `-Wl,--whole-archive` on Linux). This class of bug will recur and will always be invisible until a developer hits it locally.

## Why not reuse `mlir_wheel`

`mlir_wheel` (from eudsl) is a nightly build pinned to a rolling LLVM commit — not a stable release. While eudsl does publish `linux_aarch64` and `macosx_12_0_arm64` wheels, relying on a 30-day nightly for ARM64/macOS CI would mean the build target shifts under us and failures are hard to attribute. A stable LLVM tag gives a fixed, reproducible baseline for these runners.

## Why not build LLVM from source in CI

The original draft of this issue proposed building LLVM from source (checking out `llvm-project` at `llvmorg-22.1.0` and running a full cmake + ninja build). That approach has significant drawbacks:

- **Build time**: LLVM takes 30–60 minutes to build even with `ninja -j$(nproc)`. On GitHub-hosted macOS runners (10× billing rate), that is expensive.
- **Fragility**: The build requires platform-specific compiler flags (Homebrew Clang + libc++ on macOS vs. system Clang on Ubuntu), and any LLVM build-system change at a new tag can break the job.
- **Redundancy**: Stable LLVM releases are already available as prebuilt packages for both platforms (see below). Rebuilding them from source in CI adds no value.

## Prebuilt stable LLVM on both platforms

Both macOS and Ubuntu have well-maintained, version-pinned LLVM packages that include MLIR development files:

| Platform | Package source | MLIR packages | MLIR_DIR |
|---|---|---|---|
| Ubuntu 22.04 | `apt.llvm.org` | `libmlir-22-dev`, `mlir-22-tools` | `/usr/lib/llvm-22/lib/cmake/mlir` |
| macOS | Homebrew | `llvm` (includes MLIR by default) | `$(brew --prefix llvm)/lib/cmake/mlir` |

### Ubuntu — `apt.llvm.org`

The LLVM project publishes official apt packages at [apt.llvm.org](https://apt.llvm.org/). Setup follows the standard documented at [stephendiehl.com/posts/mlir_install](https://www.stephendiehl.com/posts/mlir_install/) and the reference Dockerfile at [github.com/sdiehl/docker-mlir-cuda](https://github.com/sdiehl/docker-mlir-cuda/blob/main/Dockerfile):

```bash
wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
add-apt-repository "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-22 main"
apt-get update
apt-get install -y libmlir-22-dev mlir-22-tools llvm-22 llvm-22-dev
```

`libmlir-22-dev` provides CMake config at `/usr/lib/llvm-22/lib/cmake/mlir/MLIRConfig.cmake`. `mlir-22-tools` provides `mlir-opt-22`, `mlir-translate-22`, etc.

### macOS — Homebrew

Homebrew's `llvm` formula builds MLIR by default (it is listed in the default `projects` list). After `brew install llvm`:

```bash
brew install llvm
echo "$(brew --prefix llvm)/bin" >> $GITHUB_PATH
```

MLIR CMake config is at `$(brew --prefix llvm)/lib/cmake/mlir/MLIRConfig.cmake`, so `MLIR_DIR` resolves to `$(brew --prefix llvm)/lib/cmake/mlir`.

## Proposed change: new `stable` job in `ci.yml`

Add a second job alongside the existing `cmake-py-test`:

```yaml
cmake-py-test-stable:
  name: cmake-py-test (stable) (${{ matrix.config.runner }})
  runs-on: ${{ matrix.config.runs-on }}

  strategy:
    fail-fast: false
    matrix:
      config:
        - { runner: "Ubuntu 22.04 x64",   runs-on: "ubuntu-22.04",     os: "ubuntu", arch: "x64"   }
        - { runner: "Ubuntu 22.04 ARM64", runs-on: "ubuntu-22.04-arm", os: "ubuntu", arch: "arm64" }
        - { runner: "macOS ARM64",        runs-on: "macos-15",          os: "macos",  arch: "arm64" }

  steps:
    - uses: actions/checkout@v4

    # ── Install prebuilt LLVM + MLIR ──────────────────────────────────────────
    # Both platforms use their native package manager to install a stable LLVM
    # release with MLIR dev files. This avoids building LLVM from source in CI.
    #
    # Ubuntu: apt.llvm.org provides versioned packages (libmlir-22-dev, etc.)
    # macOS: Homebrew llvm includes MLIR by default

    - name: Install LLVM (Ubuntu)
      if: matrix.config.os == 'ubuntu'
      run: |
        sudo apt-get update
        sudo apt-get install -y ca-certificates wget gnupg
        wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | sudo tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc
        sudo add-apt-repository "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-22 main"
        sudo apt-get update
        sudo apt-get install -y libmlir-22-dev mlir-22-tools llvm-22 llvm-22-dev cmake ninja-build

    - name: Install LLVM (macOS)
      if: matrix.config.os == 'macos'
      run: |
        brew install llvm cmake ninja python
        echo "$(brew --prefix llvm)/bin" >> $GITHUB_PATH

    # ── Resolve MLIR_DIR ──────────────────────────────────────────────────────
    # Both paths set MLIR_DIR so the remaining steps are identical.

    - name: Set paths (Ubuntu)
      if: matrix.config.os == 'ubuntu'
      run: |
        echo "MLIR_DIR=/usr/lib/llvm-22/lib/cmake/mlir" >> $GITHUB_ENV

    - name: Set paths (macOS)
      if: matrix.config.os == 'macos'
      run: |
        echo "MLIR_DIR=$(brew --prefix llvm)/lib/cmake/mlir" >> $GITHUB_ENV

    - name: Install uv
      uses: astral-sh/setup-uv@v5

    - name: Install dependencies
      run: uv sync --no-install-project --extra test

    - name: Configure ktir
      run: |
        LIT=$(uv run --no-project which lit)
        cmake -S . -B build -GNinja \
          -DCMAKE_BUILD_TYPE=Release \
          -DKTIR_ENABLE_PYTHON_BINDINGS=OFF \
          -DMLIR_DIR=${{ env.MLIR_DIR }} \
          -DLLVM_EXTERNAL_LIT=$LIT \
          -DLLVM_LIT_ARGS="-v"

    - name: Build ktir
      run: |
        JOBS=$(nproc 2>/dev/null || sysctl -n hw.logicalcpu)
        cmake --build build -j$JOBS

    - name: Test ktir
      run: cmake --build build --target check-ktir

    - name: Build and install Python wheel
      run: |
        CMAKE_ARGS="-DMLIR_DIR=${{ env.MLIR_DIR }}" uv pip install .

    - name: Python test
      run: uv run --no-project pytest -v python/test/
```

Key points:
- **No LLVM source build**: Both platforms use prebuilt packages. The `stable` job installs LLVM in ~1–2 minutes instead of 30–60 minutes.
- **Homebrew Clang on macOS**: Homebrew `llvm` provides Clang (not Apple Clang) — Apple Clang lacks flags like `-fuse-ld=lld` that may be needed. The formula's Clang is already on `PATH` after the `GITHUB_PATH` step.
- **`nproc` on macOS**: `nproc` is not available on macOS; `sysctl -n hw.logicalcpu` is the equivalent.
- **Platform linker flags**: The existing `tools/ktir-opt/CMakeLists.txt` (from PR #9) already handles `ld64` on macOS — no CMake changes needed.
- **Version pinning**: LLVM 22 is hardcoded in the workflow. See open questions about extracting this to a file.

## What is not changed

- `cmake-py-test` (Linux x64, `mlir_wheel` and `custom` paths) — untouched.
- `llvm-build.yml` — untouched.
- The `mlir-source` input and resolution logic in `ci.yml` — untouched.

## Disabling `cmake-py-test` if stable coverage is sufficient

If the team decides that `cmake-py-test (stable)` (which now includes Ubuntu x64) provides enough coverage and the nightly `mlir_wheel` / `custom` LLVM paths are no longer needed on every PR, `cmake-py-test` can be disabled by adding a top-level `if: false` condition:

```yaml
cmake-py-test:
  if: false   # disabled — coverage moved to cmake-py-test-stable
  name: cmake-py-test (${{ matrix.config.runner }})
  ...
```

`llvm-build.yml` calls `ci.yml` via `workflow_call` and passes `mlir-source: custom` — that call would then reach a skipped job, which is harmless. Alternatively, remove the `cmake-py-test` job entirely and drop the `workflow_call` trigger from `ci.yml` if `llvm-build.yml` is also retired.

`llvm-build.yml` exists solely to build a custom LLVM artifact and trigger `cmake-py-test` with `mlir-source: custom` via its `trigger-ci` job. If `cmake-py-test` is disabled or removed, `llvm-build.yml` has no useful work left and should also be disabled or deleted. The simplest approach is to remove both together.

## GitHub Actions runner cost

| Runner | Public repo | Private repo rate | Est. cost/run (private) |
|---|---|---|---|
| `ubuntu-22.04` | Free | $0.008/min (1×) | ~$0.08–$0.16 |
| `ubuntu-22.04-arm` | Free | $0.012/min (2×) | ~$0.12–$0.24 |
| `macos-15` | Free | $0.062/min (10×) | ~$0.62–$1.24 |

Cost estimates are based on ~10 minutes total (no LLVM source build) vs. ~60–90 minutes with a source build. If the repo ever becomes private, the `stable` job can be restricted to push-to-main only.

## Open questions

- [ ] LLVM 22 is the current development branch on `apt.llvm.org`. Should we pin to LLVM 20 (stable) instead? LLVM 20 is the current stable release on `apt.llvm.org`; 21 is qualification; 22 is development.
- [ ] Should the LLVM version be extracted to a file (e.g. `cmake/llvm-stable-version.txt`) rather than being hardcoded in the workflow?
- [ ] Does `libmlir-22-dev` on Ubuntu ARM64 include all the CMake config files needed by ktir? This should be verified on an `ubuntu-22.04-arm` runner.
- [ ] Does Homebrew `llvm` on `macos-15` ship `$(brew --prefix llvm)/lib/cmake/mlir/MLIRConfig.cmake`? The formula builds MLIR by default but the exact install layout should be confirmed.

## References

- [PR #9 — Prevent linker optimization affecting dialect registration](https://github.com/torch-spyre/ktir-mlir-frontend/pull/9)
- [Stephen Diehl — Installing MLIR](https://www.stephendiehl.com/posts/mlir_install/)
- [sdiehl/docker-mlir-cuda — Dockerfile](https://github.com/sdiehl/docker-mlir-cuda/blob/main/Dockerfile)
- [apt.llvm.org — LLVM apt packages](https://apt.llvm.org/)
- [GitHub Actions — About billing for GitHub Actions](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-actions/about-billing-for-github-actions)

fabianlim:
## Summary

> [!IMPORTANT]
> This PR pins LLVM to **llvmorg-22.1.3** (`e9846648`). Pre-built artifacts for Ubuntu x64, Ubuntu ARM64, and macOS ARM64 are available in this repo's GitHub Actions artifacts.

cc: @mudhakar @raghukiran1224 @kiszk @ashokponkumar @Prasanth-Chatarasi 

Closes #11

### Roadmap

- [x] **Step 1** — Wheel packaging · [PR #1](https://github.com/torch-spyre/ktir-mlir-frontend/pull/1)
- [x] **Step 2** — Stable LLVM artifact pipeline · _this PR_
- [ ] **Step 3** — Wheel publishing · [#12](https://github.com/torch-spyre/ktir-mlir-frontend/issues/12)

---

### Motivation

The end goal is to allow downstream projects (e.g. [ktir-cpu](https://github.com/torch-spyre/ktir-cpu)) to consume this repo as a dependency. To get there, we need distributable Python wheels (tracked in [#12](https://github.com/torch-spyre/ktir-mlir-frontend/issues/12)): a wheel packages the compiled ktir extensions so consumers can `pip install` without building from source.

This PR is **step 2**: pin LLVM to a stable release (llvmorg-22.1.3), produce a reproducible build artifact, and establish a clear developer flow. Without a pinned stable LLVM, wheel builds are non-reproducible and CI is fragile.

The current developer flow requires downloading the pre-built MLIR artifact (~2.4 GB on macOS ARM64) — a one-time cost that is cached locally afterward. The wheel packaging itself was already done in [PR #1](https://github.com/torch-spyre/ktir-mlir-frontend/pull/1). The remaining open question — tracked in [#12](https://github.com/torch-spyre/ktir-mlir-frontend/issues/12) — is how to publish those wheels (e.g. PyPI or a private index), so that downstream projects and developers can simply `pip install ktir-mlir-frontend` without any MLIR setup at all.

---

Reworks MLIR sourcing for CI and local development. Elevates a **pinned, self-built LLVM artifact** as the primary stable path; `mlir_wheel` becomes a bleeding-edge fallback. Introduces `scripts/setup_mlir.py` as a unified resolver used by both CI and local dev — same script, same logic everywhere.

See [docs/ci.md](docs/ci.md) for full reference.

---

### Three flows

> **Flow 1 — Standard CI** · `ci.yml` · _triggered: push/PR to `main`_
>
> The day-to-day CI run every contributor sees. Reads `cmake/llvm-hash.txt` to find the pinned LLVM version, calls `setup_mlir.py` to download the pre-built artifact, then configures, builds, and tests KTIR (LIT tests + Python tests). Falls back to `mlir_wheel` with a warning if the artifact is not found.

> **Flow 2 — LLVM Build** · `llvm-build.yml` · _triggered: hash bump in `cmake/llvm-hash.txt`, scheduled (bimonthly), or manual `workflow_dispatch`_
>
> Run when the pinned LLVM version changes or the artifact needs refreshing. Checks out `llvm/llvm-project` at the pinned commit and builds MLIR from source with `-DMLIR_ENABLE_BINDINGS_PYTHON=ON` (required for the Python wheel build in Flow 1). Detects stable release tags via `git ls-remote` — stable builds skip `check-mlir`, assertions, and unused targets to reduce build time and artifact size. Uploads the result as a 90-day GitHub Actions artifact and triggers Flow 1 against it. Skips the build entirely if a valid artifact already exists.

> **Flow 3 — Bleeding-edge** · `ci.yml` via `workflow_dispatch` with `mlir-source=mlir_wheel`
>
> Use when you want to test against the very latest LLVM `main` without going through a full LLVM build — for example, before deciding whether to bump the hash. Runs the same CI pipeline as Flow 1 but installs `mlir_wheel` from the eudsl index instead of the pinned artifact. Note: `mlir_wheel` tracks LLVM `main` and may break at any time.

---

### Changes

- **`llvm-build.yml`**: add macOS 15 ARM64 + Ubuntu ARM64; skip rebuild if artifact exists; stable/nightly build modes; `lld` on Ubuntu; 90-day retention + scheduled refresh; fix cmake flags
- **`ci.yml`**: route MLIR resolution through `setup_mlir.py`; add macOS 15 ARM64 + Ubuntu ARM64 to test matrix
- **`scripts/setup_mlir.py`** _(new)_: downloads pinned artifact (local cache), `--wheel` for mlir_wheel, `--dry-run` to check availability
- **`pyproject.toml`**: remove `mlir` optional dep; split `llvm-build-nightly` to carry `numpy<2` separately
- **`cmake/llvm-hash.txt`**: pin to `e9846648` (llvmorg-22.1.3)
- **`docs/ci.md`** _(new)_: documents the three flows and local dev workflow

---

## Testing

```bash
git clone https://github.com/torch-spyre/ktir-mlir-frontend
cd ktir-mlir-frontend && git checkout pr/llvm-build
export GIT_PAT=<your-github-pat>

# 1. Check artifact availability (no download)
uv sync --no-install-project --extra test
uv run --no-project python scripts/setup_mlir.py --dry-run

# 2. Download artifact + build Python package
MLIR_DIR=$(uv run --no-project python scripts/setup_mlir.py)
CMAKE_ARGS="-DMLIR_DIR=$MLIR_DIR" uv sync --extra test -v

# 3. Python tests
uv run --no-project pytest -v python/test/

# 4. LIT tests (separate cmake build required)
cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=Release \
  -DKTIR_ENABLE_PYTHON_BINDINGS=OFF -DMLIR_DIR=$MLIR_DIR \
  -DLLVM_EXTERNAL_LIT=$(uv run --no-project which lit)
cmake --build build && cmake --build build --target check-ktir
```

> **Fork testing:** `cmake-py-test` resolves `artifact-repo` to the parent repo. To test fork-built artifacts, use `workflow_dispatch` and pass `artifact-repo` pointing to your fork.


mudhakar:
@tinhto-000 - would appreciate your review on this - as it involved LLVM version....

scripts/setup_mlir.py
```diff
@@ -0,0 +1,354 @@
...
+) -> str:
+    """Download artifact zip, extract inner .tar.gz, unpack to cache, return MLIR_DIR."""
+    _CACHE_BASE.mkdir(parents=True, exist_ok=True)
+
+    zip_url = (
+        f"https://api.github.com/repos/{repo}/actions/artifacts/{artifact_id}/zip"
+    )
+    _err(f"Downloading {artifact_name} from {repo}...")
+    _err(f"URL: {zip_url}")
+    _err(f"Token (first 8 chars): {token[:8]}...")
```
kiszk: We should remove line 170 to reduce the security risk. Or, guard this line using an environment variable.

```
if os.environ.get("DEBUG_MLIR_SETUP"):
    _err("Token is set (length: {len(token)})")
```

kiszk (changes requested):
Address an issue regarding token's dump

scripts/setup_mlir.py
```diff
@@ -0,0 +1,354 @@
...
+
+    with tempfile.TemporaryDirectory() as _tmp:
+        tmp = pathlib.Path(_tmp)
+        zip_path = tmp / "artifact.zip"
```
kiszk, line 173: IBM Bob suggested a potential race condition when two CI jobs are executed in parallel.
Do we address this potential issue?

.github/workflows/llvm-build.yml
```diff
@@ -55,33 +56,76 @@ jobs:
...
+        if: steps.check.outputs.exists != 'true'
+        run: |
+          LLVM_HASH="${{ needs.get-hash.outputs.llvm-hash }}"
+          if git ls-remote https://github.com/llvm/llvm-project 'refs/tags/llvmorg-*^{}' \
+               | grep -q "^${LLVM_HASH}"; then
+            echo "IS_STABLE_BUILD=true"  >> $GITHUB_ENV
+          else
+            echo "IS_STABLE_BUILD=false" >> $GITHUB_ENV
+          fi
+
```
kiszk, line 111: How about the implementation thru GitHub API to reduce the elapsed time?
IBM Bob suggested the following code.

```suggestion
        run: |
          LLVM_HASH="${{ needs.get-hash.outputs.llvm-hash }}"
    
          # Query GitHub API for tags pointing to this commit
          # More efficient than git ls-remote - targeted query
          RESPONSE=$(curl -sf \
            -H "Authorization: Bearer $GH_TOKEN" \
            -H "Accept: application/vnd.github+json" \
            "https://api.github.com/repos/llvm/llvm-project/commits/$LLVM_HASH" \
            2>/dev/null || echo "{}")
    
          # Check if commit message contains release tag pattern
          if echo "$RESPONSE" | jq -r '.commit.message' | grep -q 'llvmorg-[0-9]'; then
            echo "IS_STABLE_BUILD=true" >> $GITHUB_ENV
            TAG=$(echo "$RESPONSE" | jq -r '.commit.message' | grep -o 'llvmorg-[0-9.]*' | head -1)
            echo "::notice::Detected stable release: $TAG"
          else
            # Fallback: check associated tags via separate API call
            TAGS=$(curl -sf \
              -H "Authorization: Bearer $GH_TOKEN" \
              -H "Accept: application/vnd.github+json" \
              "https://api.github.com/repos/llvm/llvm-project/commits/$LLVM_HASH/tags" \
              2>/dev/null | jq -r '.[].name' | grep '^llvmorg-' || echo "")
      
            if [ -n "$TAGS" ]; then
              echo "IS_STABLE_BUILD=true" >> $GITHUB_ENV
              echo "::notice::Detected stable release tag(s): $TAGS"
            else
              echo "IS_STABLE_BUILD=false" >> $GITHUB_ENV
              echo "::notice::No stable release tag found"
            fi
```

kiszk:
Non-requested comments are not mandatory.

.github/workflows/llvm-build.yml
fabianlim: Thanks for the suggestion! We timed `git ls-remote` with the `llvmorg-*^{}` refspec and it completes in ~0.5s against 311 tags, so there's no meaningful overhead. The current approach is also simpler and unambiguous — the GitHub API endpoint used in the suggestion (`/commits/{hash}/tags`) doesn't exist, and checking the commit message for the tag name is fragile. Happy to revisit if this becomes a bottleneck, but keeping as-is for now.

scripts/setup_mlir.py
fabianlim: The two realistic execution environments are (1) local dev on a single-user machine and (2) GitHub Actions, where each job runs in an isolated runner. In both cases concurrent writes to `_CACHE_BASE` are not a concern, so we opted not to implement locking. We have documented the safe double-checked locking pattern (check → download to tmpdir → lock → re-check → extract) in a comment in `setup_mlir.py` in case this needs to be revisited in the future.

scripts/setup_mlir.py
fabianlim: Fixed — the token logging line has been removed entirely. The URL log is sufficient for debugging without any security exposure.

kiszk (approved):

PR #21, opened 2026-05-14
Proposal: Add first cut `ktdp.reduce`: cross-core reduction op

fabianlim:
cc: @nwang-ibm @viji560 @Prasanth-Chatarasi  @vswagath1989  @tnakaike  @mudhakar @raghukiran1224 @lchu6  @moriohara

Adds a `ktdp.reduce` operation modeling cross-core reductions on
multi-core accelerators (Spyre-style). Two result-placement modes
(`all_reduce`, `reduce_to_core<N>`) — enough surface to express the
Split-K reduce-sum kernel.

## At a glance

```mlir
// Sugar form — kind set, no region.
%y = ktdp.reduce %x { kind   = #ktdp.reduce_kind<sum>,
                      mode   = #ktdp.reduce_mode<all_reduce>,
                      across = #ktdp.grid_axis<0> }
     : tensor<32x512xf16> -> tensor<32x512xf16>

// Region form — explicit combiner block + ktdp.yield, no kind.
%y = ktdp.reduce %x combiner = {
  ^bb0(%a: f16, %b: f16):
    %s = arith.addf %a, %b : f16
    ktdp.yield %s : f16
} { mode   = #ktdp.reduce_mode<reduce_to_core<0>>,
    across = #ktdp.grid_axis<0> }
  : tensor<32xf16> -> tensor<32xf16>
```

Spelling form (sugar / region) and `mode` (`all_reduce` /
`reduce_to_core<N>`) are independent — any combination is valid. The
`N` in `reduce_to_core<N>` is **rank within the reduction group**
named by `across`, not a global core ID. With `grid = [4, 8]` and
`across = grid_axis<1>`, there are 4 independent groups (one per
`pid_row`); `reduce_to_core<0>` makes 4 different rank-0 cores
writers, one per group.

## 1. Why `ktdp` needs this

Multi-core accelerators run the same SPMD program on every core; each
core holds a *partial* and the kernel needs a way to combine partials
*across cores* and say where the combined result lands. Two patterns
dominate: **replicated result** (every core needs it, feeds a
downstream op) and **single-writer result** (one core writes once to
HBM; replicating would burn ring bandwidth). `ktdp` had no way to
express either; this PR adds the op.

## 2. Design

**One overloaded op, attribute-driven.** A single `ktdp.reduce` whose
`kind` / `mode` / `across` attributes select the behavior, rather
than a family of named ops. Reductions share a shape (combiner +
result placement) that fits naturally as an attribute matrix; one op
lets a future algorithm-selection pass dispatch on `kind × mode`.

**Region is the load-bearing surface.** The combiner can be spelled
two ways: sugar (`kind = sum`, no region) and region (explicit
combiner block + `ktdp.yield`, no `kind`). Sugar is purely syntactic
— today only `sum`. The region form is what the dialect commits to
long-term, because (a) Triton's `tl.reduce` already carries an
explicit region, so region → region lowering preserves the body
unchanged; and (b) custom combiners (online-softmax-combine,
log-sum-exp, …) don't have sensible enum entries — forcing them to
grow the dialect is exactly the anti-pattern we want to avoid.

**Poison on non-writer cores.** Under `reduce_to_core<N>` every core
produces an SSA result `%y`, but only rank-`N` holds a meaningful
value; on the others `%y` is unspecified / poison. The IR is
well-formed and standard SCF passes work unchanged, but reading
non-writer results is a program bug. Kernels guard with
`scf.if pid == N`. Picked over per-core types, optional types, or
implicit guards because it keeps static IR shape identical
everywhere and matches LLVM's poison model.

## 3. Scope

In: the op (`ktdp.reduce` + `ktdp.yield`), three new attrs
(`reduce_kind`, `reduce_mode`, `grid_axis`), structural verifier,
two Split-K example kernels (one per mode).

Out, with TODOs in code: other modes (e.g. reduce-scatter);
multi-input combiners (e.g. online-softmax-combine, which reduces a
`(max, denominator, accumulator)` triple — the verifier currently
restricts to one input, lifting it is mechanical); sibling
collectives (`broadcast`, `all_gather`, `all_to_all` — not
reductions, belong in their own ops); a region → sugar canonicalizer
(waits for a second `kind` or a dispatch consumer); hardware
lowering.

## 4. Failure mode for unsupported combiners

Because the region form accepts any well-formed body, the natural
question is: *if I write a combiner the toolchain doesn't know how
to handle, when does it fail?* The contract this PR records:

- **Op verifier — structural only.** Arities, types, terminator,
  `kind`-xor-region. It does not inspect the combiner arithmetic, by
  design. Verifier guarantees well-formed IR, not runnability.
- **Lowering passes (future) must fail closed.** Any pass that
  walks `ktdp.reduce` is responsible for an explicit
  `emitOpError("…combiner not supported by this lowering pipeline…")`
  on bodies it doesn't recognize. **Silent miscompile from a
  partially-matched body is an explicit non-goal.** Documented now,
  ahead of the first lowering pass.
- **A planned recognition pass will normalize before lowering** —
  region → sugar when the body matches a known kind. Each lowering
  pass then dispatches on the enum plus a single "unrecognized
  body" path.

In short: structural verification today, fail-closed lowering
diagnostics tomorrow, recognition canonicalization between them. The
user-visible failure point is the lowering pass — never the verifier
(would force a closed enum) and never silent miscompilation.

## 5. Tests

Six new LIT files, mirroring existing dialect-wide tests one-to-one:
attrs round-trip + verify-negative (parallel to existing
`attrs-roundtrip.mlir` / `attrs-verify-negative.mlir`), op round-trip
+ invalid (parallel to `roundtrip.mlir` / `invalid.mlir`), and two
whole-kernel examples (`split-k-all-reduce.mlir`,
`split-k-reduce-to-core.mlir`) parallel to existing whole-kernel
tests like `add.mlir`. Op round-trip covers both spelling forms ×
both modes and includes an `arith.maximumf` region body to
demonstrate the open-region surface. Example kernels share shapes
so the diff between them isolates exactly the all_reduce vs.
reduce_to_core contrast. **23 / 23 LIT tests pass; 19 pre-existing.**

### DIFF
--- a/include/Ktdp/KtdpAttrs.td
+++ b/include/Ktdp/KtdpAttrs.td
@@ -100,6 +100,144 @@ def Ktdp_SpyreMemorySpaceAttr
 }
 
 //===----------------------------------------------------------------------===//
+// Reduce-related attributes (Style A from RFC_reduce_api.md)
+//===----------------------------------------------------------------------===//
+//
+// First cut: just enough surface for the split-K reduce-sum example.
+//
+//   - ReduceKindAttr  : enum of recognized combiner shorthands. Today
+//                       only `sum`. The enum is a *frontend ergonomics*
+//                       knob, NOT a capability claim — the region form
+//                       on `ktdp.reduce` accepts any combiner body.
+//
+//   - ReduceModeAttr  : where the result lands. Parametric so that
+//                       `reduce_to_core<N>` carries its destination
+//                       core inline. Today: `all_reduce` (F1) and
+//                       `reduce_to_core<N>` (F2).
+//
+//   - GridAxisAttr    : which grid axis the reduction collapses,
+//                       indexed into the kernel's `grid = [...]`
+//                       func attribute (treated as opaque here).
+//
+// TODO (deferred extensions):
+//   - ReduceModeKind: add `reduce_scatter` (F3) — carries an `axis`
+//     parameter naming the tensor dimension to scatter along.
+//   - ReduceKind: add `max`, `min`, etc. as recognized shorthands.
+//   - Other RFC primitives (`broadcast` C1, `all_gather` C2,
+//     `all_to_all` X1) belong in their own ops per the RFC's
+//     hybrid resolution, not as new entries here.
+
+// ReduceKindAttr: a parametric AttrDef wrapping the ReduceKind enum.
+// Modeled as an AttrDef (not EnumAttr) so the attribute is owned by
+// the ktdp dialect and parses with the `#ktdp.reduce_kind<...>`
+// syntax expected by the RFC. The set of entries is *frontend
+// ergonomics*, not a list of supported reductions: the region form
+// on `ktdp.reduce` accepts arbitrary combiner bodies regardless of
+// what is enumerated here. Whether a particular body lowers to a
+// hardware collective is a lowering-pass concern.
+def Ktdp_ReduceKind : I32EnumAttr<
+    "ReduceKind",
+    "Recognized combiner-shorthand kinds for ktdp.reduce",
+    [
+      I32EnumAttrCase<"sum", 0, "sum">
+      // TODO: max, min, prod, ... — add when sugar spelling is needed.
+    ]> {
+  let cppNamespace = "::mlir::ktdp";
+  let genSpecializedAttr = 0;
+}
+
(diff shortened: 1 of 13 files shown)


fabianlim:
This issue is superceed by #22 #23 #25

PR #22, opened 2026-05-15
Add design doc for inter-tile reductions (all-reduce, reduce-scatter)

Prasanth-Chatarasi:
## Summary

Adds a design document at `docs/inter-tile-reductions.md` proposing two structured operations in the `ktdp` dialect for cross-tile reductions:

- `ktdp.inter_tile_all_reduce` — every tile in a group holds that group's reduced result.
- `ktdp.inter_tile_reduce_scatter` — each group's reduced result is split row-major along a named dimension across that group's tiles; each tile holds its own slice.

Both ops share a common structure: producer / reducer regions, value-typed result semantics, identity operand, per-group participating-tile sets, and implicit synchronization via SSA dataflow over tensors (no explicit barriers in source IR). Reductions are grouped by default — disjoint subsets of tiles run independent reductions in parallel.

The doc covers:
- Operand and attribute semantics (identity, `participating_tiles_per_group`, `groups`, disjointness invariant).
- Producer region: what each tile contributes.
- Reducer region: how to combine two partials, purity requirement, unspecified ordering for scheduler freedom.
- Value-typed synchronization model.
- Result-type rules for each op (within-group tile axes collapsed; group axes preserved; scatter dimension chunked for reduce-scatter).
- Worked examples: single-group all-reduce (96x64), multi-group all-reduce (128x8x12x64), multi-group reduce-scatter (128x8x12x64).
- Notes on variadic / multi-tensor reductions, sequential composition via `scf.for`, and lowering targets.

This is a design proposal only; no code changes.

## Test plan

- [ ] Reviewers confirm the op surface and semantics are sound.
- [ ] Reviewers confirm the worked examples match the stated rules.
- [ ] Decide on remaining open questions (variadic identity ordering, enumerated-form fallback, etc.) before implementation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/inter-tile-reductions.md
+++ b/docs/inter-tile-reductions.md
@@ -1,0 +1,731 @@
+# Inter-tile Reductions in KTIR (all-reduce, reduce-scatter)
+
+**Scope:** Two structured operations in the `ktdp` dialect for cross-tile
+reductions over a partitioned set of compute tiles.
+
+---
+
+## 1. Overview
+
+`ktdp.inter_tile_all_reduce` and `ktdp.inter_tile_reduce_scatter` express
+cross-tile reductions across a participating set of compute tiles organized
+into disjoint groups. The two ops share a common structure — operand list,
+producer / reducer regions, value-typed result semantics, synchronization
+model — and differ only in what each tile receives back:
+
+- **`inter_tile_all_reduce`:** every tile in a group ends up holding that
+  group's reduced result.
+- **`inter_tile_reduce_scatter`:** each group's reduced result is split
+  row-major along a named dimension across that group's tiles; each tile
+  holds its own slice.
+
+Both ops are value-typed — their results are tensor SSA values produced
+in each participating tile's SPMD execution. There are no explicit
+barriers in the IR; synchronization arises from SSA dataflow over tensor
+values combined with the producer / reducer contract that defines what
+each tile contributes and how partials are combined.
+
+This document presents the two ops directly. Section 2 covers the
+structure shared between them. Sections 3 and 4 give per-op semantics and
+worked examples.
+
+---
+
+## 2. Common structure
+
+The pieces below apply identically to both ops. Sections 3 and 4 add only
+op-specific details (result-type rules, the scatter `dimension` attribute).
+
+### 2.1 Operands and attributes
+
+**`identity`** — variadic SSA operands, one per partial-tensor role. Each
+identity is a tensor whose **shape and element type match the
+corresponding partial type** (not the rank-reduced or scattered result
+type). The identity is the reduction's neutral element: combining it
+with any partial yields that partial. There is one identity per
+partial-tensor role, **shared across all groups and all tiles** — a
+single SSA value, hoisted before the op.
+
+**`participating_tiles_per_group`** — a parameterized affine integer set
+naming the participating tiles for a given group. The set has one
+dimension (the tile id) and one symbol (`g`, the group index). Bound at
+op-use time, the set selects the tiles that belong to group `g`. For
+example, `affine_set<(i)[g] : (i - 4*g >= 0, -i + 4*g + 3 >= 0)>` selects
+tile ids `4g .. 4g+3` for any group index `g`. An enumerated form (a
+list of tile-id lists) is supported as a fallback when the per-group
+membership is irregular and not affine-expressible.
+
+**`groups`** — an affine integer set defining the range of group indices.
+For example, `affine_set<(g) : (g >= 0, -g + 7 >= 0)>` defines 8 groups,
+indexed `0..7`.
+
+**Disjointness invariant.** For any two distinct group indices
+`g_1 != g_2` in `groups`, `participating_tiles_per_group(g_1)` and
+`participating_tiles_per_group(g_2)` must be disjoint. Equivalently,
+every participating tile is in exactly one group. The verifier enforces
+this. The motivation is unambiguous reduction membership: each tile
+contributes to exactly one group's reduction and receives exactly one
+group's result.
+
+### 2.2 Producer region
+
+The producer region indicates **what partial results each tile
+contributes to the reduction**. It is the per-tile boundary that names
+the SSA values participating in cross-tile combine. The region runs
+once per participating tile, in that tile's SPMD execution.
+
+**Region argument:** `^producer(%gid: index)` — the index of the group
+this tile belongs to. The runtime binding is direct: tile `t` finds its
+group by looking up which entry of `participating_tiles_per_group(g)`
+contains `t`; that `g` is bound to `%gid` for tile `t`'s execution of
+the producer.
+
+The body of the producer region knows its tile id via
+`ktdp.get_compute_tile_id` (the same way every SPMD KTIR body does) and
+its group index via `%gid`.
+
+**Termination:** the producer terminates with
+`ktdp.yield_partial p_1, ..., p_N`, where each `p_i` has type
+`T_partial_i` (the type of the i-th partial-tensor role declared in the
+op's signature). The yielded values may reference SSA values from the
+enclosing scope — typical use is a thin contribution marker:
+
+```mlir
+producer {
+  ^producer(%gid: index):
+    ktdp.yield_partial %my_partial : tensor<...>
+}
+```
+
+with the per-tile compute that produced `%my_partial` living at function
+scope (where it is naturally executed by every tile under SPMD). Richer
+producer bodies are allowed when the contribution is awkward to hoist —
+e.g., compute that depends only on `%gid`, or visibly local
+"contribution preparation" the author wants to keep adjacent to the op.
+
+### 2.3 Reducer region
+
+The reducer region indicates **how to perform the reduction over two
+partial results**. It is a pairwise combine — given two partials of the
+same shape (either two per-tile contributions or intermediate combined
+values), it produces a single combined value of the same shape. The
+scheduler invokes the reducer as many times as needed (in tree, ring,
+or linear order) to fold all of a group's partials into one.
+
+**Region arguments:** `^bb0(%lhs_1, ..., %lhs_N, %rhs_1, ..., %rhs_N)` —
+`2N` total, with each `lhs_i` and `rhs_i` of type `T_partial_i`.
+
+**Purity.** The reducer must be pure — no memory effects, no calls to
+side-effecting ops. Pure tensor ops (`tensor.empty`, `linalg` on
+tensors, `arith.*`) are allowed. The verifier rejects reducers that
+contain any operation with side effects.
+
+**Per-group combine.** The reducer is shared by all groups, but each
+combine only ever sees partials from tiles within the *same* group.
+Different groups' reductions are independent and may be scheduled in
+parallel.
+
+**Combine ordering.** The combine ordering is unspecified within each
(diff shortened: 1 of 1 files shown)


Prasanth-Chatarasi:
(CC @viji560 , @vswagath1989 )

docs/inter-tile-reductions.md
```diff
@@ -0,0 +1,731 @@
...
+group's result.
+
+### 2.2 Producer region
+
+The producer region indicates **what partial results each tile
+contributes to the reduction**. It is the per-tile boundary that names
+the SSA values participating in cross-tile combine. The region runs
+once per participating tile, in that tile's SPMD execution.
+
+**Region argument:** `^producer(%gid: index)` — the index of the group
```
fabianlim, line 77: This sounds like an interesting idea, im having some difficulty coming up with a non-trival use case. Beacuse producers are limited to only "map" kind of transformers, and the knob here is `gid`, im having some difficulty thinking why this cannot be done before coming to the reduce call

docs/inter-tile-reductions.md
```diff
@@ -0,0 +1,731 @@
...
+**Termination:** the producer terminates with
+`ktdp.yield_partial p_1, ..., p_N`, where each `p_i` has type
+`T_partial_i` (the type of the i-th partial-tensor role declared in the
+op's signature). The yielded values may reference SSA values from the
+enclosing scope — typical use is a thin contribution marker:
+
+```mlir
+producer {
+  ^producer(%gid: index):
+    ktdp.yield_partial %my_partial : tensor<...>
```
fabianlim, line 96: this is just a minor nit, but just wondering why we might not want to just go for a generic `ktdp.yield`. Why do we want to differentiate for `partial` cases?

docs/inter-tile-reductions.md
```diff
@@ -0,0 +1,731 @@
...
+
+```mlir
+#A_view_set  = affine_set<(d0, d1) : (d0 >= 0, -d0 + 95 >= 0, d1 >= 0, -d1 + 63 >= 0)>
+#AB_tile_set = affine_set<(d0, d1) : (d0 >= 0, -d0 +  2 >= 0, d1 >= 0, -d1 + 63 >= 0)>
+#E_view_set  = affine_set<(d0, d1) : (d0 == 0, d1 >= 0, -d1 + 63 >= 0)>
+#E_tile_set  = affine_set<(d0, d1) : (d0 == 0, d1 >= 0, -d1 + 63 >= 0)>
+#identity_2d = affine_map<(d0, d1) -> (d0, d1)>
+
+// One group containing all 32 tiles.
+#group_tiles = affine_set<(i)[g] : (i - 32*g >= 0, -i + 32*(g+1) - 1 >= 0)>
```
fabianlim, line 232: Using affine_sets is a great idea, though was wondering if it might make sense to use a sugar for. There is some benefit to using something like ktdp.grid_axis<0>, so we can infer the layouts of the groups from the grid =[x,y] attribute. On the other hand, writing the group size 32 may introduce inconsistencies between group_tiles and the grid.
Alternatively, writing it like this affine_set<(i) : (i >= 0, -i + 31 >= 0) may be  simpler if we only have one group.

An actually we can avoid the inconsistency issue by introducing a symbol for the group size also affine_set<(i)[N] : (i >= 0, -i + N - 1 >= 0)>

docs/inter-tile-reductions.md
```diff
@@ -0,0 +1,731 @@
...
+                    output_shape [128, 1, 1, 64]
+                    : tensor<128x1x64xf16> into tensor<128x1x1x64xf16>
+
+    // Multi-group all-reduce. Within each group of 4 tiles, dim 2 of the
+    // partial is the within-group tile axis (collapsed). Dim 1 is the
+    // group axis (preserved). Each tile gets its group's reduced <128x1x64>.
+    %my_group_result = ktdp.inter_tile_all_reduce
+        identity(%add_id : tensor<128x1x64xf16>)
+        participating_tiles_per_group = #group_tiles,
+        groups = #all_groups
```
fabianlim, line 456: Having a reduce to core semantic can potentially be more efficient if the result is not needed on all cores. could we consider adding something like `leader_tile_per_group`, which if for consistentcy we have it to be an affine_set also. Then one might write to express the leaders

```
leader_tile_per_group= affine_set<(i)[g] : (i == 4*g)>     // first tile of each group of 4`
leader_tile_per_group = affine_set<(i)[g] : (i == 4*g + 3)> // last tile of each group of 4
```

Prasanth-Chatarasi: Yes, we can consider introducing a new operation `ktdp.inter_tile_single_reduce` taking owner (core/tile) information.

docs/inter-tile-reductions.md
Prasanth-Chatarasi: We can certainly introduce shorthand forms or simplified constructs. 

In the group tiles shown above, “i” represents the axis-0 of the grid. In the example, since we have a 1-D grid of tiles, you’ll see “i”. If we have a 2D grid, you’ll start seeing “(I,j)” in the input space of affine sets.

Prasanth-Chatarasi: > An actually we can avoid the inconsistency issue by introducing a symbol for the group size also affine_set<(i)[N] : (i >= 0, -i + N - 1 >= 0)>

In the group_tiles, g is a symbolic parameter in the affine set.

docs/inter-tile-reductions.md
Prasanth-Chatarasi: Ah, it’s often a common practice to use a yield operation that’s specific to a particular region. This simplifies validation and downstream passes. For instance, if there’s a common yield operation, you’d need to compare its parent region to all possible region types to obtain the correct base pointer. On the other hand, if it’s a specific yield operation, it’s guaranteed to be part of that particular region.

docs/inter-tile-reductions.md
Prasanth-Chatarasi: `gid` is exposed for future use, in case indexing needs to be done using `group_id` instead of `tile_id`. Computing `group_id` from `tile_id` can introduce arithmetic operations that could be avoided using this block argument.

docs/inter-tile-reductions.md
fabianlim: Thanks for clarifying, makes sense.

docs/inter-tile-reductions.md
fabianlim: > In the group_tiles, g is a symbolic parameter in the affine set.

Ah yes i missed that, yes we are on the same page.

docs/inter-tile-reductions.md
fabianlim: > gid is exposed for future use, in case indexing needs to be done using group_id instead of tile_id.

Yes I understand, but would having the producers region, be much improved over having a small preamble like in the examples; if we could replace the producer region with an argument for example:

```
// get some offset from ktdp.get_compute_tile_id : index
%partial_2d = ... compute partial from offset

// pass partial as an argument
%reduced = ktdp.inter_tile_all_reduce %partial_2d
        identity(%add_id : tensor<64xf16>)
        participating_tiles_per_group = #group_tiles,
        groups = #all_groups
        : (tensor<1x64xf16>) -> tensor<64xf16>
    reducer {
      ^bb0(%lhs: tensor<1x64xf16>, %rhs: tensor<1x64xf16>):
        %init = tensor.empty() : tensor<1x64xf16>
        %sum  = linalg.add ins(%lhs, %rhs : tensor<1x64xf16>, tensor<1x64xf16>)
```

Also:
1. most use cases will load something specific to that group, but if we think of offloading that logic to inside the producer region, it may look unwieldly
2. I cant seem to think of another use case of the gid, other than to pick a tile offset to load from.

docs/inter-tile-reductions.md
```diff
@@ -0,0 +1,731 @@
...
+                    ins(%AB_sum : tensor<3x64xf16>)
+                    outs(%red_filled : tensor<64xf16>)
+                    dimensions = [0]
+    %partial_2d = tensor.expand_shape %partial_1d [[0, 1]] output_shape [1, 64]
+                    : tensor<64xf16> into tensor<1x64xf16>
+
+    // Single-group all-reduce. The unit dim 0 is the within-group tile axis;
+    // the op collapses it. Every tile holds the same %reduced.
+    %reduced = ktdp.inter_tile_all_reduce
+        identity(%add_id : tensor<64xf16>)
```
fabianlim, line 289: Sorry i forgot to ask this. I understand that `identity` is intended to be the identity of the partial
- do you mean the identity under the `reducer` operation? The identity element could be different depending on the operation (e.g., add vs mult)
- Perhaps there is a typo here, the type should be `tensor<1x64xf16>`?

I understand that this identity is different from the `init` in the `reducer` region, but it does seem to make the ktir slightly verbose. Perhaps in most cases they may actually be the same identity, so if there is somehow a way to reuse across them .. .

docs/inter-tile-reductions.md
Prasanth-Chatarasi: > most use cases will load something specific to that group, but if we think of offloading that logic to inside the producer region, it may look unwieldly

You don't have to offload that logic to the producer region, hence you see examples where the logic is in fact outside the producer region. The producer region emphasizes on what partial results from participating tiles should go into reduction. I understand that eliminating the producer region can simplify the operation in its syntax (by couple of lines), but for completeness, it's good to have a producer region with yield operands passing the partial results. For, example, think about a "hypothetical scenario" where you want to do reduction with partial results from participating tiles and non-reduction with non-participating tiles. 

> I cant seem to think of another use case of the gid, other than to pick a tile offset to load from.

Agreed, gid is also added for completeness in the sense that if you want to access something based on gid, its better to have a block argument coming from producer region instead of explicitly computing via arithmetic operations.

docs/inter-tile-reductions.md
Prasanth-Chatarasi: > do you mean the identity under the reducer operation? The identity element could be different depending on the operation (e.g., add vs mult)

Yes, it is identity element under a reduction operation. For, example. it would be zero tensor in sum reduction, and negative infinity for max reduction. The concept is similar to linalg.reduce's init arguments. 

> Perhaps there is a typo here, the type should be tensor<1x64xf16>

Similar to linalg.reduce semantics, the output tensor or identity tensor of reduction operation is rank reduced tensor, hence you are seeing <64xf16> instead of <1x64xf16>

docs/inter-tile-reductions.md
fabianlim: >  For, example, think about a "hypothetical scenario" where you want to do reduction with partial results from participating tiles and non-reduction with non-participating tiles.

Sorry just to make sure I am not misunderstanding your thinking, can I clarify this? 
1. every tile has to belong to some group `g`?
2. for every group `g`, every tile in that group must commits to the reduce.
3. the producer region allows a transformation on the partial, but has no ability to preclude a tile from not participating in the reduce.

Unless you are thinking of the scenario where every tile from the non-participating group, will `ktdp.yield` the reduction's neutral element (i.e. the identity)?

fabianlim: > but for completeness, it's good to have a producer region with yield operands passing the partial results.

Also in this respect if we do consider the region for completeness sake, would we consider allowing the `providers` region as an *optional* region?

PR #23, opened 2026-05-20
Inter-tile communications in KTIR

tnakaike:
This is a PR to extend a proposal for #22 .

Extension summary

- Separate producer and consumer regions so that we can express various communication patterns between producer and consumer tiles.
- Introduce ktdp.tile_future to express an in-flight tile that needs to be materialized by a consumer (delivery) op before using it for computation.
- Introduce ktdp.inter_tile_consume as a consumer op for broadcast without reduction semantics


docs/inter-tile-communication.md
```diff
@@ -0,0 +1,913 @@
...
+### 5.2 Reducer region
+
+Identical to §4.2: pure, associative-commutative, combine ordering
+unspecified. Different groups' reductions proceed independently.
+
+### 5.3 Type rules
+
+`T_r` is `T_p` with the within-group tile axes collapsed and then sliced
+along `scatter_dimension` (same as the `scatter_dimension`-present case
+in `general-inter-tile.md` §2.3).
```
fabianlim: stray reference to a local .md file.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,913 @@
...
+---
+
+## 8. Pattern instantiation
+
+### 8.1 Broadcast  →  `inter_tile_produce` + `inter_tile_consume`
+
+```mlir
+// Tile 0 loads W; future carries it to all 4 tiles.
+%W_future = ktdp.inter_tile_produce
+    producer_tiles_per_group = #tile_0,          // {i | i == 4*g}
```
fabianlim: Understand this is for 4 tiles, but the notation `{i | i == 4*g}` was a little bit confusing to me. I guess you want to say

```
  #tile_0       = affine_set<(i) : (i == 0)>             // just tile 0
  #all_group_tiles = affine_set<(i) : (i >= 0, -i + 3 >= 0)>  // tiles 0..3
  #single_group = affine_set<(g) : (g == 0)>             // one group
```

here it gets confusing to me because we lost `participating_tiles_per_group`, which was clear to be mecause
- groups: specified the range of `g`
- participating_tiles_per_group: for a given `g`, what are the tiles

but here `producer_tiles_per_group` does not have a symbol `g`. So it makes `groups` feel very arbitrary. The same comment on `inter_tile_consume`

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,913 @@
...
+### 2.3 Op signature
+
+```mlir
+%future = ktdp.inter_tile_produce
+    producer_tiles_per_group = <affine-set>,
+    groups                   = <affine-set>
+    : T_p -> !ktdp.tile_future<T_p>
+{
+  ^bb0(%gid: index):
+    ktdp.yield_partial %val : T_p
```
fabianlim: the proposal in #22 did also mention multi-arity reductions, so then we might want to consider multi-arity `yield_partial`'s

tnakaike:
Fixed the document to reflect @fabianlim 's comments.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
+# Inter-tile communications in KTIR
+
+**Scope:** Four ops — `ktdp.inter_tile_produce`, `ktdp.inter_tile_consume`,
```
Prasanth-Chatarasi, line 3: @tnakaike - I appreciate your efforts in taking this direction towards separating the production and consumption parts of the communication. This helps beyond reduction, i.e., it can support in interacting with other tiles without materializing into the memory and access them via the `construct_distributed_memory_view`.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
...
+2. **Delivery** — how the contributed data is delivered to the receiving
+   tiles: pass-through unchanged (broadcast), folded by a combiner
+   (reduce), or folded then scattered (reduce-scatter).
+
+Separating these concerns into a unified production op plus three
+delivery ops keeps each op single-purpose and enables any combination:
+
+| Pattern | Production op | Delivery op |
+|---------|--------------|-------------|
+| Broadcast | `ktdp.inter_tile_produce` | `ktdp.inter_tile_consume` |
```
Prasanth-Chatarasi, line 30: I highlight and categorize these operations as work-sharing constructs of SPMD programming model (similar to `omp for` or `omp sections` in OpenMP model).

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
...
+| Reduce | `ktdp.inter_tile_produce` | `ktdp.inter_tile_reduce` |
+| Reduce-scatter | `ktdp.inter_tile_produce` | `ktdp.inter_tile_reduce_scatter` |
+
+`ktdp.inter_tile_produce` returns a `!ktdp.tile_future<T_p>` SSA value.
```
Prasanth-Chatarasi: Futures are typically synchronization handlers in task-based parallel programming models, but it's quite interesting and useful to see this in the SPMD context.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
...
+production to delivery encodes the happens-before ordering with no
+explicit barriers in the IR.
+
+---
+
+## 2. `ktdp.inter_tile_produce` — unified production op
+
+### 2.1 Attributes
+
+**`producer_tiles_per_group`** — parameterized affine integer set `(i)[g]`
```
Prasanth-Chatarasi, line 55: I assume you are inheriting the concept of groups from the base PR. This is because groups can be multi-dimensional in addition to tiles being multi-dimensional.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
...
+`affine_set<(i)[g] : (i - 4*g >= 0, -i + 4*g + 3 >= 0)>` selects tile ids
+`4g .. 4g+3` for any group index `g`. An enumerated form (a list of tile-id
+lists) is supported as a fallback when per-group membership is irregular.
+- Broadcast: selects exactly one tile per group.
+- Reduce / reduce-scatter: selects all tiles in the group.
+
+**Disjointness invariant.** For any two distinct group indices `g_1 != g_2`
+in `groups`, `producer_tiles_per_group(g_1)` and
+`producer_tiles_per_group(g_2)` must be disjoint. Every producing tile is
+in exactly one group. The verifier enforces this. The motivation is
```
Prasanth-Chatarasi, line 67: How do you plan to enforce this in verifier (without enumeration)?

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
...
+    producer_tiles_per_group = <affine-set>,
+    groups                   = <affine-set>
+    : T_p_1, ..., T_p_N -> !ktdp.tile_future<T_p_1, ..., T_p_N>
+{
+  ^bb0(%gid: index):
+    ktdp.yield_partial %val_1, ..., %val_N : T_p_1, ..., T_p_N
+}
+```
+
+`%future` is a workgroup-visible handle. It cannot materialize until all
```
Prasanth-Chatarasi: This is very restrictive blocking semantics, i.e., all producers tiles irrespective of groups should wait for all other tiles to finish. Is that really needed?

On a separate note, I'm not sure if this is sufficient because consider a scenario where there is one producer and there is one consumer which is a different one (assume core0 and core1). What enforces core1 to wait for core0 to finish producing?

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
...
+
+---
+
+## 3. `ktdp.inter_tile_consume` — plain delivery op (broadcast)
+
+### 3.1 Operand and attributes
+
+**Operand:** `!ktdp.tile_future<T_p>` — the future returned by the
+corresponding `ktdp.inter_tile_produce`. The def-use edge is the ordering
+constraint.
```
Prasanth-Chatarasi: "The def-use edge is the ordering constraint." -- I think this is the key. You may want to say that all consumer tiles participating in the `inter_tile_consume_*` operations should wait for all producer tiles to finish because of future construct.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
...
+
+**Operand:** `!ktdp.tile_future<T_p>` returned by
+`ktdp.inter_tile_produce`.
+
+**`consumer_tiles_per_group`** — tiles that receive the reduced result.
+When equal to `producer_tiles_per_group`, every contributing tile receives
+the result (all-reduce). When it is a strict subset of the producer set,
+two additional patterns are expressible:
+
+- **Reduce-to-one** (`consumer_tiles_per_group` = {1 tile}): all N tiles
```
Prasanth-Chatarasi: I don't think these two variations may not be needed. We could say that all participating tiles in consumer region gets all reduced values.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
...
+### 4.5 Result semantics
+
+The op produces N variadic SSA values, one per partial-tensor role. The
+values are *per-tile-valued* — each consumer tile holds a result value
+when the op completes. Every consumer tile in a group holds the same
+value — that group's fully reduced result. Tiles in different groups hold
+different values (each its own group's reduction).
+
+**Non-participating tiles.** A tile not in `consumer_tiles_per_group` for
+any group must not lexically encounter the op in its execution path. When
```
Prasanth-Chatarasi: "must not" --> this can make IR bloated with conditionals on the tile id's or group id's. Perhaps, how about  saying that the results are undefined for tiles not participating in that consumer/reducer region.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
...
+**Non-participating tiles.** Same constraint as §4.5.
+
+---
+
+## 6. Synchronization model
+
+No explicit barriers appear in the IR. The `!ktdp.tile_future<T_p>` SSA
+value encodes two ordering constraints:
+
+1. `%future` cannot materialize until all producer tiles in each group
```
Prasanth-Chatarasi: I'm wondering if this restriction is focused per group instead of all groups?

Prasanth-Chatarasi (commented):
Overall, the proposal looks positive. Added my comments.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,932 @@
...
+all-reduce, and reduce-scatter.
+
+---
+
+## 1. Motivation
+
+Inter-tile communication involves two orthogonal concerns:
+
+1. **Production** — which tiles contribute data and what they contribute.
+2. **Delivery** — how the contributed data is delivered to the receiving
```
Prasanth-Chatarasi, line 15: To help with the backend scheduler, I suggest to put a restriction on the producer return values (futures) and consumers (reduction/scatter/..) to be in the same basic block.

docs/inter-tile-communication.md
tnakaike: Yes. This inherits your proposal.

docs/inter-tile-communication.md
tnakaike: I did not have the plan when I wrote this proposal since this is [a copy of your proposal](https://github.com/torch-spyre/ktir-mlir-frontend/blob/80978cdcf683537469554b152945774c3893429e/docs/inter-tile-reductions.md#21-operands-and-attributes).

My current idea is to use a SMT solver which will allow us to verify if a memory area for a group overlaps with another memory area for another group.

@airin711 Is this verification possible by using a SMT solver? You studied this technique to verify the equivalence between before- and after- lowering IRs.

docs/inter-tile-communication.md
tnakaike: Right. I also think that this is too restrictive. We may need a per-tile (or per-group) sync handle. Let me think more flexible expressions.

> On a separate note, I'm not sure if this is sufficient because consider a scenario where there is one producer and there is one consumer which is a different one (assume core0 and core1). What enforces core1 to wait for core0 to finish producing?

When a tile already resides in a scratchpad of core 0, the producer block should have ktdp.load from the the scratchpad. Core 1 waits for core 0 to load the tile from the scratchpad.

docs/inter-tile-communication.md
tnakaike: I see. Let me fix it.

docs/inter-tile-communication.md
tnakaike: I see. Let me fix it.

docs/inter-tile-communication.md
tnakaike: Right. Let me add the restriction.

tnakaike:
@Prasanth-Chatarasi Thank you so much for your comments. I fixed the doc to reflect your comments except for the future semantics (per-tile or per-group?).

docs/inter-tile-communication.md
airin711: Yes, an SMT-based verifier can model the memory region assigned to each possible group ID as a corresponding tile within its internal memory model, and verify that these tiles do not overlap with one another without enumerating the memory regions.

docs/inter-tile-communication.md
fabianlim: @Prasanth-Chatarasi that sounds interesting. Do you have an example how `inter_tile_produce` may interact with say `construct_distributed_memory_view`?

docs/inter-tile-communication.md
fabianlim: @Prasanth-Chatarasi it seems this is only about enumerating groups being disjoint, these are very small affine sets, sorry for the naive question, but why cant we verify disjointness without enumeration? Am I missing something?

@airin711 @tnakaike it seems that this is about groups being disjoint. Im a little confused why we are bringing memory area into the picture. Yes while its true maybe thats a worthwhile validation as well, but is it required?

tnakaike:
@Prasanth-Chatarasi @fabianlim I updated the document to enable a per-tile future (synchronization handle).

- Add an attribute producer_dependency_per_consumer for each delivery op. This attribute defines a mapping from a producer tile and a delivery (consumer) tile to avoid waiting for all producer tiles to be complete.
- Add a restriction for a future so that it can be used only by a single delivery op. This simplifies the verification to avoid dead locks.
- Add possible extensions for gather and scatter patterns.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,1249 @@
...
+```mlir
+%future = ktdp.inter_tile_produce
+    producer_tiles_per_group = <affine-set>,
+    groups                   = <affine-set>
+    : T_p_1, ..., T_p_N -> !ktdp.tile_future<T_p_1, ..., T_p_N>
+{
+  ^bb0(%gid: index):
+    ktdp.yield_partial %val_1, ..., %val_N : T_p_1, ..., T_p_N
+}
+```
```
fabianlim, line 124: the full-barrier / per-tile discussion forward-references an
attribute that's only defined in §3.1. Could it move to §6?

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,1249 @@
...
+**`producer_dependency_per_consumer`** *(optional)* — affine integer set
+`(p)[c, g]` over producer tile IDs `p`, parameterized by consumer tile
+`c` and group index `g`. For consumer tile `c` in group `g`, only the
+producer tiles satisfying this set are waited on and received by the
+delivery op. If absent, the consumer waits for all producer tiles in the
+group (full-barrier semantics).
+
+**Verifier invariants when present:**
+
+1. **Subset check.** The declared set must be a subset of
```
fabianlim, line 159: Writing `P`, `C`, `D` for the three sets, could the
invariants be stated more precisely as:

- subset:   `D[c,g] ⊆ P[g]` for every `c ∈ C[g]`
- coverage: `P[g] ⊆ ⋃_{c ∈ C[g]} D[c,g]`

Together these give `P[g] = ⋃_{c ∈ C[g]} D[c,g]`, which I think is
the intent?

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,1249 @@
...
+   `producer_tiles_per_group`. Referencing a non-producer tile is a
+   verifier error.
+2. **Coverage check.** For every group `g` and every producer tile `p`
+   in `producer_tiles_per_group(g)`, at least one consumer tile `c` in
+   `consumer_tiles_per_group(g)` must satisfy
+   `producer_dependency_per_consumer(p)[c, g]`. A producer tile absent
+   from the union is a verifier error — it would yield a value that no
+   consumer ever reads, risking a deadlock in push-based lowerings.
+
+Because a `%future` has exactly one delivery op (§2.3), these invariants
```
fabianlim, line 179: Just confirm that this requires two checks?

1. produce side — `getResult().getUses()` count is exactly one
   (rejects both zero uses, i.e. orphan producer, and >1 uses).
2. delivery side — `getFuture().getDefiningOp()` is an
   `inter_tile_produce` (rejects deliveries fed by anything else
   returning `!ktdp.tile_future<T>`, e.g. function args).

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,1249 @@
...
+
+---
+
+## 8. Pattern instantiation
+
+### 8.1 Broadcast  →  `inter_tile_produce` + `inter_tile_consume`
+
+```mlir
+// 4 tiles, 1 group: tile 0 loads W; all 4 tiles compute.
+#tile_0          = affine_set<(i)[g] : (i - 4*g == 0)>
```
fabianlim, line 432: can I confirm this example means 

```
#tile_0(i) = {i==0}
#all_group_tiles(i)[g==0] = {0,1,2,3}
```

So basically this is a broadcast restricted to 4 tiles

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,1249 @@
...
+assembled output, where `chunk` is the size of `T_p` along
+`gather_dimension`.
+
+**Ordering.** Unlike the reduce combiner — which may be applied in any
+order — the placement of each producer's partial in the output is
+determined by its within-group local index. The ordering is
+deterministic and does not require commutativity.
+
+**Relationship to other ops.** Gather is the inverse of
+`inter_tile_reduce_scatter`: reduce-scatter reduces N partials then
```
fabianlim: This made me double-take, as for the the natural inverse of gather is `scatter` and not `reverse-scatter`. Since you already discuss scatter below, maybe its more natural to point to that instead?

docs/inter-tile-communication.md
tnakaike: Right. We want to avoid dead locks where a producer waits for a consumer to receive a tile forever.

docs/inter-tile-communication.md
tnakaike: Right.

docs/inter-tile-communication.md
tnakaike: Good catch. This is redundant and brigs confusions. I removed the related ops sections, and put a summary of op coverage as a table though this is not complete. Let's focus on per-tile synchronizations first, and the extend it to other operation patterns.

tnakaike:
@fabianlim Thank you so much for your comments. I reflected your comments.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,1243 @@
...
+        : tensor<1x64xf16> -> !ktdp.tile_future<tensor<1x64xf16>>
+    {
+      ^bb0(%gid: index):
+        ktdp.yield_partial %partial_2d : tensor<1x64xf16>
+    }
+
+    // Reduce: unit dim 0 is the within-group tile axis; the op collapses it.
+    // Every tile holds the same %reduced : tensor<64xf16> (all-reduce case).
+    %reduced = ktdp.inter_tile_reduce(%partial_future)
+        consumer_tiles_per_group = #group_tiles,
```
fabianlim, line 572: @tnakaike so for `inter_tile_reduce`, for every `g`, we must always have `consumer_tiles_per_group[g] == producer_tiles_per_group[g]`? Otherwise Im confused what will be the `lhs` and `rhts` of the combiner region in the reducer. 

And if that is true, do we want to remove `consumer_tiles_per_group` in the reduce?

docs/inter-tile-communication.md
fabianlim: @tnakaike just to confirm, your intent is different from mine above

  Construct a counterexample. Suppose two groups, g ∈ {0, 1}:

  - P[0] = {0, 1}, P[1] = {2, 3}
  - C[0] = {4}, C[1] = {5}
  - D[4, 0] = {3} — consumer 4 (in group 0) declares dependency on tile 3.

  My formulation rejects this: D[4, 0] = {3} and P[0] = {0, 1}, so {3} ⊄ {0, 1}. Verifier error — consumer in group 0 references a tile that isn't a
  producer in group 0.

  Yours accepts this: {3} ⊆ ⋃_g P[g] = {0, 1, 2, 3}. Tile 3 is a producer somewhere (in P[1]), so the global subset check passes.

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,1243 @@
...
+{
+  ^bb0(%gid: index):
+    %W = ktdp.load ...
+    ktdp.yield_partial %W : tensor<64x128xf16>
+}
+
+// Every consumer tile extracts its copy; no combiner → value passes through.
+%W_tile = ktdp.inter_tile_consume(%W_future)
+    consumer_tiles_per_group = #all_group_tiles,
+    groups                   = #single_group
```
fabianlim: @tnakaike just to confirm, if `producer_tiles_per_group` has more than 1 tile, then which tile's value will pass through to `inter_tile_consume`?

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,1243 @@
...
+  ^bb0(%gid: index):
+    %data = ktdp.load ...
+    ktdp.yield_partial %data : tensor<64xf16>
+}
+
+// Each consumer unblocks independently as its assigned producer finishes.
+%my_data = ktdp.inter_tile_consume(%data_future)
+    consumer_tiles_per_group         = #consumer_tiles,
+    groups                           = #single_group,
+    producer_dependency_per_consumer = #dep_per_consumer
```
fabianlim, line 999: @tnakaike also a similar question as above, if `producer_dependency_per_consumer` allows multiple tiles, such as

```
  #single_group     = affine_set<(g) : (g == 0)>
  #producer_tiles   = affine_set<(i)[g] : (i - 1 >= 0, -i + 2 >= 0)>
  #consumer_tiles   = affine_set<(i)[g] : (i - 3 >= 0, -i + 4 >= 0)>
  #dep_per_consumer = affine_set<(p)[c, g] : (p - 1 >= 0, -p + 2 >= 0)>
```

In the above example, my both tile 3 and 4 will depend on both producer tiles {1,2}, then whose partials will be passed through

docs/inter-tile-communication.md
tnakaike: No. I do not want to remove `consumer_tiles_per_group`. Please note that this is a result set. When we specify a single tile in `consumer_tiles_per_group`, we can get a single reduce tile (not all-reduce case). You also pointed out that we want to do it.

`lhs` and `rhs` are **abstract** partial tensors. Implementation should be given by the backend. Typically, `rhs` is a producer tile.

docs/inter-tile-communication.md
tnakaike: It cannot be handled. This is a typo. Let me fix it.

docs/inter-tile-communication.md
tnakaike: Maybe, we use multiple tiles in two semantcis.

1. Multiple tiles yielded from a single ktdp.inter_tile_produce
2. Multiple tiles defined in producer_tiles_per_group and consumer_tiles_per_group

They are orthogonal. Multiple tiles in 1 are handled as a single tile (a single set of tiles) in producer_tiles_per_group and consumer_tiles_per_group. I also think that this is confusing. I updated the specification to yield multiple tensors by reflecting [your comment](https://github.com/torch-spyre/ktir-mlir-frontend/pull/23#discussion_r3270977842).

Should we remove this specification? Maybe, the same thing can be done by using multiple sets of producer and consumer.

docs/inter-tile-communication.md
fabianlim: ok so are you saying that 
1. if `consumer_tiles_per_group[g]` is *not* a single, tile, then we expect `consumer_tiles_per_group == producer_tiles_per_group`
2. if `consumer_tiles_per_group[g]` is a single tile `c`, then we can have `c \in producer_tiles_per_group[g]`

So 1 is the canonical all-reduce use case, and 2 is the cononical single-reduce use case.

tnakaike: There is another pattern. Originally, we had the following statement. But, I simplified it based on @Prasanth-Chatarasi 's [comment](https://github.com/torch-spyre/ktir-mlir-frontend/pull/23#discussion_r3295774039).

```
consumer_tiles_per_group — tiles that receive the reduced result. When equal to producer_tiles_per_group, every contributing tile receives the result (all-reduce). When it is a strict subset of the producer set, two additional patterns are expressible:

Reduce-to-one (consumer_tiles_per_group = {1 tile}): all N tiles contribute partials but only one tile receives the result. Useful when one tile aggregates the output for a downstream stage (e.g., writing to memory) while the remaining tiles proceed independently.
Reduce-to-subset (1 < |consumer set| < N): an M-tile subset of the contributing tiles receives the result. The N − M tiles outside the consumer set are not required to wait for or hold the result.
```

docs/inter-tile-communication.md
fabianlim: > Should we remove this specification? Maybe, the same thing can be done by using multiple sets of producer and consumer.

For `inter_tile_consume`, if `producer_tiles_per_group[g]` are multiple, then its well defined only if `producer_dependency_per_consumer` can resolve to a single producer per consumer

and yes if we want a consumer to consumer from say 2 producers, then we repeat it twice, its verbose but it cant be helped, because currently we are not able to express the the order of the producer's yields, if a single consumer tile is to receive from multiple producers.

> Multiple tiles yielded from a single ktdp.inter_tile_produce

I think you slightly confound the terms here. Im using the term `tile` as how you are using it, which I believe to mean a "core". I think you use it to mean core, and also to mean a "chunk of a tensor"

docs/inter-tile-communication.md
```diff
@@ -0,0 +1,1243 @@
...
+### 3.1 Operand and attributes
+
+**Operand:** `!ktdp.tile_future<T_p_1, ..., T_p_N>` — the future returned
+by the corresponding `ktdp.inter_tile_produce`. The def-use edge is the
+ordering constraint.
+
+**`consumer_tiles_per_group`** — tiles that receive the delivered value
+per group.
+
+**`groups`** — must match the corresponding `inter_tile_produce`.
```
bmahjour: Since the groups need to match between producer/consumer they have to be redundantly specified on both ops. Wouldn't it be better to fold this information into the proposed `ktdp.tile_future` type, since it acts as the link between producer/consumers?

eg: `!ktdp.tile_future<T, #group>`

that way both producer/delivery ops can infer the group info from the result/operand type rather than re-declaring it, and any mismatches would be caught by the type system instead of relying on the verifier.

tnakaike:
@bmahjour @fabianlim I reflected Bardia's comments.

PR #25, opened 2026-06-08
KTDP inter-tile communication: all-reduce + single-reduce (produce + reduce)

fabianlim:
Implements the all-reduce pattern from `docs/inter-tile-communication.md` (PR #23).
The doc lands four ops; this PR lands the two needed for all-reduce plus the
shared scaffolding (future type, terminators) the other patterns will reuse.

## 1. What is implemented (and where the rest goes)

**Implemented:**
- Type `!ktdp.tile_future<T_p_1, ...>` — variadic partial roles (`KtdpTypes.{td,cpp}`).
- Terminators `ktdp.yield_partial`, `ktdp.yield_reduced` (`KtdpOps.td`).
- `ktdp.inter_tile_produce` — producer region, `producer_tiles_per_group`/`groups`
  affine sets, single-use future result.
- `ktdp.inter_tile_reduce` — combiner region (2N args), `identity` operands,
  optional `producer_dependency_per_consumer`, collapsed-axis result.
  Verifiers in `KtdpOps.cpp`.

**Reduce variants — supported modes and the spec relationship.** The verifier
checks `consumer_tiles_per_group` (C) against `producer_tiles_per_group` (P) per
group `g`. The spec (§4.1, §4.5, open question Q1) and our policy differ — this
is a deliberate, documented restriction, **not** spec enforcement:

| Relation per group `g` | Spec status | This impl |
|---|---|---|
| `C == P` | mandated valid (all-reduce) | **accept** |
| `\|C\| == 1`, `C ⊆ P` | supported (reduce-to-one) | **accept** |
| `1 < \|C\| < \|P\|`, `C ⊆ P` | **spec-supported** (reduce-to-subset) | **reject — "unsupported"** (our restriction; IR is spec-legal) |
| `C ⊄ P` (a consumer that did not produce) | **open (Q1)** | **reject — "unsupported (Q1)"** |

The hard, spec-mandated rule is `C ⊆ P`; everything tighter (only all-reduce +
reduce-to-one) is our choice. Error messages say **"unsupported"**, never
"invalid", so we never misrepresent the spec. (`groups` must-match is a true
spec rule — still TODO, see §2.)

**Future patterns slot into the same files, no new scaffolding:**
- Broadcast → new `ktdp.inter_tile_consume` op in `KtdpOps.{td,cpp}` (consumes
  the existing future, no combiner).
- Reduce-scatter → new `ktdp.inter_tile_reduce_scatter` op, = reduce + a
  `scatter_dimension` attr + sliced result type.
- Additional consumers per future / per-tile sync → relaxing the single-use
  rule (§2) + `producer_dependency_per_consumer` subset/coverage checks, all in
  `InterTileReduceOp::verify()` / the new consume op's verifier.
- Gather, scatter → new ops, same pattern.

## 2. Polyhedral checks — status and where they go

Implemented via enumeration assuming that upper bounds are present; remaining checks have TODO markers.

| Check | Lives in | Status |
|---|---|---|
| Producer group disjointness (§2.1) | `InterTileProduceOp::verify` | ✅ done (enumerated) |
| Consumer-vs-producer: `C ⊆ P` + supported-mode gate | `InterTileReduceOp::verify` | ✅ done (enumerated) |
| `groups` matches between produce & reduce (§4.1) | `InterTileReduceOp::verify` | ✅ done (enumerated) |
| `producer_dependency_per_consumer` subset + coverage (§3.1, §6) | `InterTileReduceOp::verify` | ✅ done (enumerated) |
| Symbolic variants (unbounded `groups` range) of the two done checks | both verifiers | NOT DOING |

**Enumerated approach (all done checks).** When the `groups` range is statically
bounded, pin each concrete `g` via `FlatLinearValueConstraints(IntegerSet)` +
`setAndEliminate`, then enumerate tile ids into `DenseSet<int64_t>` and check by
set membership — no Presburger set-operation queries. New CMake dependency:
`MLIRAnalysis` (not `MLIRPresburger`).
- **Disjointness:** for each concrete `g`, enumerate `tilesOf(producerSet, g)`
  into a `DenseSet`; check pairwise intersection across groups by membership
  lookup.
- **Consumer-vs-producer:** enumerate `tilesOf(consumerSet, g)` and
  `tilesOf(producerSet, g)`; check `C ⊆ P` by membership, then accept iff
  `C == P` (all-reduce) or `|C| == 1` (reduce-to-one).
- **`groups`-match:** direct `IntegerSet` value equality on the two attrs.
- **Dependency subset/coverage:** enumerate `depTilesOf(depSet, c, g)` and
  verify each `p` is in `P(g)` (subset), then verify every `p ∈ P(g)` is
  covered by some consumer's dependency set (coverage).

A pure-Presburger alternative (Option A) was also designed — disjointness and
subset become relation-emptiness queries over `(i, g)` without any loop; the
`|C|==1` mode gate requires a two-tile-variable construction. See the PR
discussion for the full design if unbounded-groups support is ever needed.

When `groups` is **not** statically bounded, all checks are skipped with a
`// TODO(inter-tile): symbolic` marker (a symbolic-`g` emptiness query is the
follow-up).

**Single-use invariants — both now covered:**
- ✅ **Future → one consumer:** `InterTileProduceOp::verify` enforces the future
  has exactly one use.
- ✅ **Consumer → one producer (in effect):** the consumer-vs-producer check
  reaches the producer via `getFuture().getDefiningOp<InterTileProduceOp>()`; a
  future not defined by a single `inter_tile_produce` makes the cross-op
  comparison unavailable (the check defers). Combined with the future single-use
  rule, a well-formed reduce consumes exactly one producer's future.

## 3. Tests — added, covered, and gaps

**Added** (`test/Ktdp/`):
- `tile-future-type.mlir` — type round-trip (single + multi-role).
- `inter-tile-reduce-roundtrip.mlir` — single-group, multi-group (preserved
  group axis), **reduce-to-one** (`|C|==1`, accepted), and N=2 argmax with a
  `producer_dependency_per_consumer` set. The argmax case uses a real 2-adic
  combiner (`cmpf ogt` / `maxnumf` / `select` — the same body `tt.reduce`
  argmax lowers to), not a passthrough, so it exercises an actual value+index
  correlated reduction.
- `inter-tile-reduce-invalid.mlir` — 9 negatives: yield arity, future
  multi-use, identity-type mismatch, collapse rank, non-unit collapse,
  non-tensor future partial, **producer-group non-disjointness** (§2.1),
  **consumer not subset of producer** (Q1, unsupported), and
  **reduce-to-subset** (spec-legal but unsupported).

Both full published examples (§8.2.1 96×64, §8.2.2 128×8×12×64) round-trip
through `ktir-opt` unchanged.

**Covers:** type shape, region arity, collapsed-axis rule, identity match,
future single-use, variadic (argmax) structure, producer-group disjointness,
and the full consumer-vs-producer mode matrix (accept all-reduce + reduce-to-one;
reject reduce-to-subset and `C⊄P`).

**Test gaps (track with §2):**
- No negatives for `groups` mismatch or dependency-set subset/coverage (those
  checks are still TODO).
- No symbolic (unbounded-`groups`) tests — both done checks defer there.
- No use-by-non-consumer-tile (§4.5 ownership, open question Q4).

## Spec-compliance summary

All-reduce and reduce-to-one: **validated** (`C ⊆ P` mandated check + mode gate).
Reduce-to-subset and `C ⊄ P` (Q1): **rejected as unsupported** — a deliberate
restriction, surfaced with "unsupported" messages, not spec violations.
Broadcast, reduce-scatter, per-tile sync, gather, scatter: **not in this PR.**
Remaining spec checks (`groups` match, dependency subset/coverage) and the
symbolic-`groups` variants are documented TODOs at known sites.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,13 @@ The `ktdp` dialect models tile-based, data-parallel kernels targeting multi-core
 - **Data movement**: `load`, `store` (driven by access tiles)
 - **Tile identity**: `get_compute_tile_id`
 - **Runtime args**: `runtime_arg` type + `runtime_arg_extract` op
+- **Inter-tile communication**: `inter_tile_produce` + `inter_tile_reduce`
+  (all-reduce; `yield_partial` / `yield_reduced` terminators). See
+  [docs/inter-tile-communication.md](docs/inter-tile-communication.md). Broadcast
+  (`inter_tile_consume`), reduce-scatter, per-tile sync, gather, and scatter are
+  future work.
 
-Custom types: `!ktdp.access_tile<NxMxindex>`, `!ktdp.runtime_arg<type, granularity=N, upperbound=M>`
+Custom types: `!ktdp.access_tile<NxMxindex>`, `!ktdp.runtime_arg<type, granularity=N, upperbound=M>`, `!ktdp.tile_future<T_p_1, ...>`
 
 Memory spaces: `#ktdp.spyre_memory_space<HBM>`, `<LX, core=7>`, `<unspecified>`
 
--- a/include/Ktdp/KtdpOps.td
+++ b/include/Ktdp/KtdpOps.td
@@ -73,6 +73,164 @@ def KTDP_RegionTerminatorOp : Ktdp_Op<"region_terminator", [Terminator]> {
   let assemblyFormat = "attr-dict";
 }
 
+//===----------------------------------------------------------------------===//
+// Inter-tile communication
+//
+// Implements the design in docs/inter-tile-communication.md. This first cut
+// covers the all-reduce pattern: `ktdp.inter_tile_produce` (production) +
+// `ktdp.inter_tile_reduce` (reduction delivery), full-barrier synchronization.
+// Broadcast (`inter_tile_consume`), reduce-scatter, per-tile synchronization
+// (`producer_dependency_per_consumer`), gather, and scatter are future work;
+// the `!ktdp.tile_future` type and these terminators are shared scaffolding for
+// them.
+//===----------------------------------------------------------------------===//
+
+def KTDP_YieldPartialOp : Ktdp_Op<"yield_partial",
+    [Terminator, ReturnLike, Pure,
+     HasParent<"::mlir::ktdp::InterTileProduceOp">]> {
+  let summary = "Terminator naming a tile's partial contributions";
+  let description = [{
+    Terminates the producer region of `ktdp.inter_tile_produce`, yielding one
+    value per partial-tensor role. The yielded value types must match the
+    partial types carried by the produced `!ktdp.tile_future`.
+  }];
+  let arguments = (ins Variadic<AnyTensor>:$values);
+  let assemblyFormat = "($values^ `:` type($values))? attr-dict";
+}
+
+def KTDP_YieldReducedOp : Ktdp_Op<"yield_reduced",
+    [Terminator, ReturnLike, Pure,
+     ParentOneOf<["::mlir::ktdp::InterTileReduceOp"]>]> {
+  let summary = "Terminator yielding the combined partials of a reducer region";
+  let description = [{
+    Terminates the reducer region of `ktdp.inter_tile_reduce` (and, in future,
+    `ktdp.inter_tile_reduce_scatter`), yielding one combined value per
+    partial-tensor role. The yielded value types must match the partial types
+    `T_p_i` (not the rank-reduced result types).
+  }];
+  let arguments = (ins Variadic<AnyTensor>:$values);
+  let assemblyFormat = "($values^ `:` type($values))? attr-dict";
+}
+
+def KTDP_InterTileProduceOp : Ktdp_Op<"inter_tile_produce",
+    [SingleBlockImplicitTerminator<"::mlir::ktdp::YieldPartialOp">,
+     RecursiveMemoryEffects]> {
+  let summary = "Unified production op for inter-tile communication";
+  let description = [{
+    The `ktdp.inter_tile_produce` operation names **what partial results each
+    tile contributes** to a cross-tile communication and returns a
+    `!ktdp.tile_future` handle carrying per-tile availability signals.
+
+    The producer region runs once per participating tile in that tile's SPMD
+    execution. Its single block argument `%gid : index` is the index of the
+    group the tile belongs to (the runtime binds tile `t` to the group `g`
+    whose `producer_tiles_per_group(g)` contains `t`). The block terminates
+    with `ktdp.yield_partial`, naming the per-tile contribution for each role.
+
+    Attributes:
+    - `producer_tiles_per_group` — parameterized affine integer set `(i)[g]`
+      selecting which tiles produce per group.
+    - `groups` — affine integer set defining the range of valid group indices.
+
+    The result `!ktdp.tile_future` must have exactly one use: the single
+    delivery op that consumes it.
+
+    Example:
+    ```mlir
+    %f = ktdp.inter_tile_produce
+        producer_tiles_per_group = affine_set<(i)[g] : (i - 4*g >= 0, -i + 4*g + 3 >= 0)>,
+        groups                   = affine_set<(g) : (g >= 0, -g + 7 >= 0)>
+        : tensor<1x64xf16> -> !ktdp.tile_future<tensor<1x64xf16>>
+    {
+      ^bb0(%gid: index):
+        ktdp.yield_partial %partial : tensor<1x64xf16>
+    }
+    ```
+  }];
+
+  let arguments = (ins
+    Builtin_IntegerSetAttr:$producer_tiles_per_group,
+    Builtin_IntegerSetAttr:$groups);
+  let results = (outs Ktdp_TileFutureType:$future);
+  let regions = (region SizedRegion<1>:$body);
+
+  let hasCustomAssemblyFormat = 1;
+  let hasVerifier = 1;
+
+  let extraClassDeclaration = [{
+    /// Convenience: the partial types carried by the produced future.
+    ::llvm::ArrayRef<::mlir::Type> getPartialTypes() {
+      return ::llvm::cast<TileFutureType>(getFuture().getType())
+          .getPartialTypes();
+    }
+  }];
+}
+
+def KTDP_InterTileReduceOp : Ktdp_Op<"inter_tile_reduce",
+    [SingleBlockImplicitTerminator<"::mlir::ktdp::YieldReducedOp">,
+     RecursiveMemoryEffects]> {
+  let summary = "Reduction delivery op (all-reduce / reduce)";
+  let description = [{
+    The `ktdp.inter_tile_reduce` operation consumes the `!ktdp.tile_future`
+    produced by `ktdp.inter_tile_produce` and combines the partials of every
+    producer tile in each group with an associative-commutative combiner,
+    delivering the fully reduced result to the consumer tiles.
+
+    The reducer region receives `2N` block arguments
+    (`%lhs_1..%lhs_N, %rhs_1..%rhs_N`, each of partial type `T_p_i`) and
+    terminates with `ktdp.yield_reduced`, yielding `N` combined values of the
+    partial types. The combiner must be pure.
+
+    Attributes:
+    - `consumer_tiles_per_group` — tiles that receive the reduced result.
+    - `groups` — must match the corresponding `inter_tile_produce`.
+    - `producer_dependency_per_consumer` *(optional)* — per-tile dependency
+      set; when absent, full-barrier semantics (all producers contribute).
+
+    `identity` provides one identity tensor per role (matching `T_p_i`),
+    hoisted before the op and shared across groups/tiles.
+
+    Each result `T_r_i` is `T_p_i` with the within-group tile axes collapsed.
+    Every consumer tile in a group holds that group's fully reduced result
+    (all-reduce when consumer set == producer set).
+
+    Example:
+    ```mlir
+    %r = ktdp.inter_tile_reduce(%f)
+        consumer_tiles_per_group = affine_set<(i)[g] : (i - 4*g >= 0, -i + 4*g + 3 >= 0)>,
+        groups                   = affine_set<(g) : (g >= 0, -g + 7 >= 0)>,
+        identity(%add_id : tensor<1x64xf16>)
+        : !ktdp.tile_future<tensor<1x64xf16>> -> tensor<64xf16>
+    {
+      ^bb0(%lhs: tensor<1x64xf16>, %rhs: tensor<1x64xf16>):
+        %s = linalg.add ins(%lhs, %rhs : tensor<1x64xf16>, tensor<1x64xf16>)
+                        outs(%lhs : tensor<1x64xf16>) -> tensor<1x64xf16>
+        ktdp.yield_reduced %s : tensor<1x64xf16>
+    }
+    ```
+  }];
+
+  let arguments = (ins
+    Ktdp_TileFutureType:$future,
+    Variadic<AnyTensor>:$identity,
+    Builtin_IntegerSetAttr:$consumer_tiles_per_group,
+    Builtin_IntegerSetAttr:$groups,
+    OptionalAttr<Builtin_IntegerSetAttr>:$producer_dependency_per_consumer);
+  let results = (outs Variadic<AnyTensor>:$results);
+  let regions = (region SizedRegion<1>:$combiner);
+
+  let hasCustomAssemblyFormat = 1;
+  let hasVerifier = 1;
+
+  let extraClassDeclaration = [{
+    /// Convenience: the partial types carried by the consumed future.
+    ::llvm::ArrayRef<::mlir::Type> getPartialTypes() {
+      return ::llvm::cast<TileFutureType>(getFuture().getType())
+          .getPartialTypes();
+    }
+  }];
+}
+
 def KTDP_GetComputeTileIdOp : Ktdp_Op<"get_compute_tile_id",
   [Pure]> {
     let summary = "Gets the id of the current compute tile";
--- a/include/Ktdp/KtdpTypes.td
+++ b/include/Ktdp/KtdpTypes.td
@@ -213,4 +213,58 @@ def Ktdp_RuntimeArgType : KTDP_Type<"RuntimeArg", "runtime_arg"> {
   let genVerifyDecl = 1;
 }
 
+//===----------------------------------------------------------------------===//
+// TileFutureType
+//===----------------------------------------------------------------------===//
+
+def Ktdp_TileFutureType : KTDP_Type<"TileFuture", "tile_future"> {
+  let summary = "Workgroup-visible handle to in-flight per-tile partial contributions";
+  let hasCustomAssemblyFormat = 1;
+  let description = [{
+    Syntax:
+
+    ```
+    tile_future-type ::= `tile_future` `<` type (`,` type)* `>`
+    ```
+
+    The `tile_future` type is the SSA handle returned by
+    `ktdp.inter_tile_produce`. It carries one or more *partial-tensor roles*
+    (the variadic `T_p_1, ..., T_p_N`), each a ranked tensor describing what a
+    producer tile contributes to the cross-tile communication. The def-use edge
+    from the producing op to the single consuming delivery op
+    (`inter_tile_consume` / `inter_tile_reduce` / `inter_tile_reduce_scatter`)
+    encodes the happens-before ordering with no explicit barriers in the IR.
+
+    N >= 1 roles are supported; N >= 2 expresses correlated tuples such as the
+    (values, indices) pair of an argmax-style reduction.
+
+    Examples:
+
+    ```mlir
+    // Single partial role.
+    !ktdp.tile_future<tensor<1x64xf16>>
+
+    // Two correlated roles (argmax-style: values, indices).
+    !ktdp.tile_future<tensor<128xf32>, tensor<128xi32>>
+    ```
+  }];
+
+  let parameters = (ins ArrayRefParameter<"::mlir::Type">:$partialTypes);
+
+  let builders = [
+    TypeBuilderWithInferredContext<(ins
+      "::llvm::ArrayRef<::mlir::Type>":$partialTypes
+    ), [{
+      return $_get(partialTypes.front().getContext(), partialTypes);
+    }]>
+  ];
+
+  let extraClassDeclaration = [{
+    /// Number of partial-tensor roles carried by this future.
+    size_t getNumPartials() const { return getPartialTypes().size(); }
+  }];
+
+  let genVerifyDecl = 1;
+}
+
 #endif // KTDP_TYPES_TD
--- a/lib/Ktdp/KtdpDialect.cpp
+++ b/lib/Ktdp/KtdpDialect.cpp
@@ -42,7 +42,8 @@ void KtdpDialect::initialize() {
   >();
   addTypes<
     AccessTileType,
-    RuntimeArgType
+    RuntimeArgType,
+    TileFutureType
   >();
   addAttributes<
     SpyreMemorySpaceAttr
--- a/lib/Ktdp/KtdpOps.cpp
+++ b/lib/Ktdp/KtdpOps.cpp
@@ -22,6 +22,7 @@
 #include "Ktdp/KtdpOps.hpp"
 
 #include "Ktdp/KtdpTypes.hpp"
+#include "llvm/ADT/DenseSet.h"
 #include "llvm/ADT/TypeSwitch.h"
 #include "mlir/Analysis/FlatLinearValueConstraints.h"
 #include "mlir/Analysis/Presburger/IntegerRelation.h"
@@ -32,7 +33,6 @@
 
 using namespace mlir;
 using namespace mlir::ktdp;
-using presburger::IntegerPolyhedron;
 
 //===----------------------------------------------------------------------===//
 // ConstructAccessTilesOp
@@ -1229,38 +1229,89 @@ LogicalResult RuntimeArgExtractOp::verify() {
 
 namespace {
 
-// Build the polyhedron of tile ids selected by a parameterized tile set for a
-// single concrete group value `g`, over the tile-id dimension. `tileSet` is a
-// set `(i)[g]` (e.g. producer_tiles_per_group or consumer_tiles_per_group); we
-// materialize it as constraints and pin its lone symbol `g` to `gVal`, leaving
-// an IntegerPolyhedron over `i`.
-IntegerPolyhedron tilesForGroup(IntegerSet tileSet, int64_t gVal) {
-  FlatLinearValueConstraints cst(tileSet);
-  // The symbol introduced by the integer set is the group index `g`. Fix it to
-  // the concrete value and eliminate it, leaving the tile-id dimension(s).
-  unsigned symOffset = cst.getVarKindOffset(presburger::VarKind::Symbol);
-  cst.setAndEliminate(symOffset, {gVal});
-  return IntegerPolyhedron(cst);
-}
+// Pure-enumeration helpers for inter-tile verifiers (Option B).
+//
+// We enumerate concrete integer points rather than using Presburger set
+// operations. This keeps the code simple and removes the MLIRPresburger
+// dependency. The trade-off is that both `groups` and the tile-id range must
+// be statically bounded; unbounded cases are deferred (TODO).
+//
+// Note: a pure-Presburger approach (Option A) was also designed: disjointness
+// and subset become relation emptiness queries over (i, g) without any loop;
+// the |C|==1 mode gate requires a two-tile-variable construction. See the
+// discussion at https://github.com/torch-spyre/ktir-mlir-frontend/pull/25
+// for the full design if unbounded-groups support is ever needed.
 
-// Concrete group values that satisfy `groupsSet`, when its range is statically
-// bounded. Returns std::nullopt if unbounded (caller should defer the check).
-std::optional<SmallVector<int64_t>> boundedGroupValues(IntegerSet groupsSet) {
-  FlatLinearValueConstraints groupsCst(groupsSet);
+// Enumerate the group values in `groupsSet` (a 1D set over `g` with no
+// symbols). Returns std::nullopt when the range is not statically bounded.
+std::optional<SmallVector<int64_t>> groupValues(IntegerSet groupsSet) {
+  FlatLinearValueConstraints cst(groupsSet);
+  // getConstantBound64 returns a scalar constant lb/ub or nullopt when the
+  // bound is symbolic or absent -- exactly the bounded-vs-unbounded gate we
+  // need, with no manual constraint inspection.
   std::optional<int64_t> lo =
-      groupsCst.getConstantBound64(presburger::BoundType::LB, /*pos=*/0);
+      cst.getConstantBound64(presburger::BoundType::LB, /*pos=*/0);
   std::optional<int64_t> hi =
-      groupsCst.getConstantBound64(presburger::BoundType::UB, /*pos=*/0);
+      cst.getConstantBound64(presburger::BoundType::UB, /*pos=*/0);
   if (!lo || !hi) return std::nullopt;
   SmallVector<int64_t> vals;
   for (int64_t g = *lo; g <= *hi; ++g) {
     FlatLinearValueConstraints gCst(groupsSet);
-    IntegerPolyhedron gPoly(gCst);
-    gPoly.setAndEliminate(gPoly.getVarKindOffset(presburger::VarKind::SetDim),
-                          {g});
-    if (!gPoly.isIntegerEmpty()) vals.push_back(g);
+    gCst.setAndEliminate(gCst.getVarKindOffset(presburger::VarKind::SetDim),
+                         {g});
+    if (!gCst.isIntegerEmpty()) vals.push_back(g);
   }
   return vals;
+}
+
+// Return the set of tile ids selected by `tileSet` (a set `(i)[g]`) for a
+// concrete group value `g`. Enumerates `i` from 0 up to the static upper
+// bound on the tile-id dimension. Returns std::nullopt when that bound is not
+// statically known.
+std::optional<llvm::DenseSet<int64_t>> tilesOf(IntegerSet tileSet,
+                                               int64_t gVal) {
+  FlatLinearValueConstraints cst(tileSet);
+  // Fix the group symbol to gVal and project it out.
+  cst.setAndEliminate(cst.getVarKindOffset(presburger::VarKind::Symbol),
+                      {gVal});
+  // Upper bound on the tile-id dimension after fixing g; nullopt means the
+  // tile range is symbolic (e.g. parameterized by a runtime value) -- defer.
+  std::optional<int64_t> hi =
+      cst.getConstantBound64(presburger::BoundType::UB, /*pos=*/0);
+  if (!hi) return std::nullopt;
+  llvm::DenseSet<int64_t> out;
+  for (int64_t i = 0; i <= *hi; ++i) {
+    FlatLinearValueConstraints pt(tileSet);
+    pt.setAndEliminate(pt.getVarKindOffset(presburger::VarKind::Symbol),
+                       {gVal});
+    pt.setAndEliminate(pt.getVarKindOffset(presburger::VarKind::SetDim), {i});
+    if (!pt.isIntegerEmpty()) out.insert(i);
+  }
+  return out;
+}
+
+// Return the set of producer tiles in `depSet` (a set `(p)[c, g]`) for a
+// concrete consumer `cVal` and group `gVal`. Symbols are ordered [c, g].
+std::optional<llvm::DenseSet<int64_t>> depTilesOf(IntegerSet depSet,
+                                                   int64_t cVal,
+                                                   int64_t gVal) {
+  FlatLinearValueConstraints cst(depSet);
+  unsigned symBase = cst.getVarKindOffset(presburger::VarKind::Symbol);
+  cst.setAndEliminate(symBase, {cVal});   // fix c (first symbol)
+  cst.setAndEliminate(symBase, {gVal});   // fix g (now first remaining symbol)
+  std::optional<int64_t> hi =
+      cst.getConstantBound64(presburger::BoundType::UB, /*pos=*/0);
+  if (!hi) return std::nullopt;
+  llvm::DenseSet<int64_t> out;
+  for (int64_t p = 0; p <= *hi; ++p) {
+    FlatLinearValueConstraints pt(depSet);
+    unsigned sb = pt.getVarKindOffset(presburger::VarKind::Symbol);
+    pt.setAndEliminate(sb, {cVal});
+    pt.setAndEliminate(sb, {gVal});
+    pt.setAndEliminate(pt.getVarKindOffset(presburger::VarKind::SetDim), {p});
+    if (!pt.isIntegerEmpty()) out.insert(p);
+  }
+  return out;
 }
 
 }  // namespace
@@ -1350,22 +1401,27 @@ LogicalResult InterTileProduceOp::verify() {
                        "(the group index g)");
 
   // Disjointness invariant (§2.1): producer tile sets for distinct groups must
-  // not overlap. Enumerated check -- only when the group range is statically
-  // bounded; otherwise defer (TODO: symbolic emptiness over g1 != g2).
-  if (auto groupVals = boundedGroupValues(groupsSet)) {
-    for (size_t a = 0; a < groupVals->size(); ++a)
-      for (size_t b = a + 1; b < groupVals->size(); ++b) {
-        IntegerPolyhedron pa = tilesForGroup(producerSet, (*groupVals)[a]);
-        IntegerPolyhedron pb = tilesForGroup(producerSet, (*groupVals)[b]);
-        if (!pa.intersect(pb).isIntegerEmpty())
-          return emitOpError("producer_tiles_per_group for groups ")
-                 << (*groupVals)[a] << " and " << (*groupVals)[b]
-                 << " are not disjoint";
-      }
+  // not overlap. Enumeration check -- requires statically bounded groups and
+  // tile-id range; deferred otherwise.
+  if (auto groupVals = groupValues(groupsSet)) {
+    SmallVector<llvm::DenseSet<int64_t>> tileSets;
+    for (int64_t g : *groupVals) {
+      auto ts = tilesOf(producerSet, g);
+      if (!ts) break;  // tile bound not static for this g -- skip
+      tileSets.push_back(std::move(*ts));
+    }
+    if (tileSets.size() == groupVals->size()) {
+      for (size_t a = 0; a < tileSets.size(); ++a)
+        for (size_t b = a + 1; b < tileSets.size(); ++b)
+          for (int64_t tile : tileSets[a])
+            if (tileSets[b].count(tile))
+              return emitOpError("producer_tiles_per_group for groups ")
+                     << (*groupVals)[a] << " and " << (*groupVals)[b]
+                     << " are not disjoint";
+    }
   }
-  // else: unbounded group range -- defer to a future symbolic check.
+  // else: unbounded -- defer to a future symbolic check.
 
-  // TODO(inter-tile): symbolic disjointness check for unbounded group ranges.
   return success();
 }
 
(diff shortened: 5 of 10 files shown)


fabianlim:
Thanks for @Prasanth-Chatarasi to pointing out the wierd mixture of Presburger and enumeration in the current implementation. Here is just a think-through of going either full Presburger or full enumeration:

I feel simple is best and we just go for enumeration, any thoughts @tnakaike 


# Inter-tile verifier: pure-Presburger vs pure-enumeration

**Context.** The current `InterTileProduceOp` / `InterTileReduceOp` verifiers use
a *hybrid*: they enumerate concrete group values (`boundedGroupValues` loops `g`
from lo..hi) and then run Presburger ops (`isSubsetOf`, `intersect`,
`computeVolume`) per `g`. Review feedback: this is the worst of both worlds —
we pay for Presburger machinery while remaining fundamentally an enumeration
that breaks on unbounded `groups`. Pick a lane:

- **A — pure Presburger:** evaluate nothing (not even `g`); prove each property
  as a set/relation query over the whole parameterized family at once.
- **B — pure enumeration:** drop Presburger entirely; enumerate the integer
  points of every set into plain containers and use set arithmetic.

The checks at stake (per the spec + our policy):

1. **Disjointness (§2.1):** for all `g₁ ≠ g₂`, `P(g₁) ∩ P(g₂) = ∅`.
2. **Subset (mandated):** `C(g) ⊆ P(g)` for **every** `g`.
3. **Mode gate (our policy):** for **every** `g`, accept `C(g) == P(g)`
   (all-reduce) or `|C(g)| == 1` (reduce-to-one); reject reduce-to-subset and
   `C ⊄ P`.

All conditions are **∀ `g` ∈ `groups`** — they must hold in *every* group, not
just some. This quantifier matters for the encoding below (see the mode gate).

### Notation

`producer_tiles_per_group` and `consumer_tiles_per_group` are affine integer
sets `(i)[g]` whose **dimension is the tile id `i`** and whose **symbol is the
group index `g`**. We read them two ways; the doc uses both, so they are defined
here once:

- **Predicate form `P(i, g)` / `C(i, g)`** — the boolean "is tile `i` a
  producer (resp. consumer) in group `g`?", i.e. the point `(i, g)` satisfies
  the set's affine constraints. Example:
  `#cons = affine_set<(i)[g] : (i - 4*g >= 0, -i + 4*g + 3 >= 0)>` means
  `C(i, g) = (4g ≤ i ≤ 4g+3)`, so `C(5, 1)` is true and `C(5, 0)` is false.
  Keeping *both* `i` and `g` as variables gives the full **relation** over
  `(i, g)` — this is what Option A (pure Presburger) operates on, so one
  `isSubsetOf` covers all groups at once.
- **Set-of-tiles form `P(g)` / `C(g)`** — *fix* `g`, then
  `P(g) = { i : P(i, g) }` is the set of producer tile ids in that one group.
  This is what Option B (enumeration) materializes per group as a
  `DenseSet<int64_t>`.

`groups` is an affine set over `g` alone (no `i`, no symbols). Throughout,
`P`/`C` abbreviate the two sets; `T` (Option B only) is an upper bound on tile
ids.

---



## Option A — pure Presburger (evaluate nothing)

Keep `g` symbolic throughout. No `groups` enumeration, no bound check; works for
unbounded `groups`.

### Disjointness

"No tile `i` is a producer in two different groups", i.e. the relation
`{ (i, g₁, g₂) : P(i, g₁) ∧ P(i, g₂) ∧ g₁ ≠ g₂ }` is empty. Built from two
copies of `P` sharing the tile var `i` but with distinct group symbols `g₁`,
`g₂`; the `g₁ ≠ g₂` is split into `g₁ < g₂` and `g₂ < g₁`.

```cpp
// Pseudocode over IntegerRelation. P1 = P(i, g1), P2 = P(i, g2).
IntegerRelation rel = P1;
rel.mergeAndAlignSymbols(P2);     // shared i, distinct g1, g2
rel.append(P2);                   // P(i, g1) AND P(i, g2)
// g1 < g2  (the < case; then the symmetric g2 < g1 case)
rel.addInequality(/* g2 - g1 - 1 >= 0 */);
if (!rel.isIntegerEmpty())
  return emitOpError("producer tile sets for distinct groups overlap");
```

No loop. One relation, one (well, two for the split) emptiness check.

### Subset `C ⊆ P`

`C` and `P` as relations over `(i, g)` (same `groups` constraints appended);
one call:

```cpp
if (!C.isSubsetOf(P))
  return emitOpError("consumer not subset of producer (Q1 unsupported)");
```

`isSubsetOf` quantifies over all `g` automatically. `C == P` is `C.isEqual(P)`.

### Mode gate `∀g : |C(g)| == 1`

The fiddly one. "Exactly one per group, for every group" = **at-least-one per
group** ∧ **at-most-one per group**, and the two directions encode differently
because the requirement is `∀g`:

```cpp
// AT MOST ONE (∀g): no group contains two distinct consumers, i.e. the relation
// { (i, j, g) : C(i, g) ∧ C(j, g) ∧ i < j } is empty. One emptiness check gives
// the ∀g, since "empty" means "no such (i,j,g) triple in ANY group".
//
// We build it from two copies of C that must (a) SHARE the symbol g but
// (b) use DISTINCT tile variables i and j. The two ops below do different jobs:
//   - mergeAndAlignSymbols: reconcile the SYMBOL spaces so both copies' `g`
//     refer to the SAME symbol column -- this is what enforces "same group".
//   - append: add the other copy's CONSTRAINTS, giving C(i,g) ∧ C(j,g).
// NOTE: getting "distinct tile vars" is the fiddly part -- a plain second copy
// reuses dimension 0, collapsing j onto i. The real code must insertVar a
// second tile dimension and place Cj's tile constraints on it (insertVar /
// convertVarKind), so the two lines below are schematic, not literal.
IntegerRelation twoPts = Ci;       // C over (i, g)
twoPts.mergeAndAlignSymbols(Cj);   // align symbol spaces: the two `g` become one
twoPts.append(Cj);                 // conjoin Cj's constraints: C(i,g) ∧ C(j,g)
twoPts.addInequality(/* j - i - 1 >= 0 */);   // i < j  (distinctness via order)
bool atMostOne = twoPts.isIntegerEmpty();

// AT LEAST ONE (∀g): every group in `groups` has some consumer. NOT plain
// !C.isIntegerEmpty() -- that only means SOME group is nonempty. The ∀g version
// is a coverage check, expressed (as usual in Presburger) by asking whether a
// COUNTEREXAMPLE exists: a group that should have a consumer but doesn't.
//
//   getRangeSet() is a symbolic projection (eliminates i), NOT enumeration.
PresburgerSet gWithConsumer = C.getRangeSet();   // { g : ∃i. C(i, g) } -- groups that DO have a consumer
// uncovered = groups that SHOULD have a consumer (in `groups`) but DON'T.
// This is the set of violating groups; "at least one per group" holds iff there
// are none. (Same idiom as subset: groups ⊆ gWithConsumer  <=>  groups \ gWithConsumer == ∅.)
PresburgerSet uncovered = PresburgerSet(groups).subtract(gWithConsumer);
bool atLeastOne = uncovered.isIntegerEmpty();   // empty violation set => covered everywhere

bool reduceToOne = atMostOne && atLeastOne;
```

The "at least one" direction is the subtle one: bare non-emptiness answers
*some* `g`, but the mode gate (like every condition here) is **∀g**, so it needs
the `groups ⊆ proj_g(C)` coverage check.

**Verdict.** Clean for disjointness/subset/`C==P`. The `|C|==1` gate needs a
second tile-var copy + disequality split. ~complete (handles unbounded groups,
strided sets). Cost is conceptual: `mergeAndAlignSymbols` / `append` /
`convertVarKind` plumbing and getting the splits right.

---

## Option B — pure enumeration (evaluate all affine sets)

Drop Presburger. Require `groups` (and the tile range) bounded; enumerate every
integer point into `DenseSet<int64_t>`s and use set arithmetic. Affine-set
membership is a direct constraint evaluation at concrete `(i, g)` — no
Presburger, no `IntegerRelation`.

```cpp
// Needs: bounded group range [g0,g1] and a tile-id upper bound T.
auto tilesOf = [&](IntegerSet s, int64_t g) {
  llvm::DenseSet<int64_t> out;
  for (int64_t i = 0; i < T; ++i)
    if (s.isConstraintSatisfied(/*dims=*/{i}, /*syms=*/{g})) out.insert(i);
  return out;
};

for (int64_t g = g0; g <= g1; ++g) {
  auto Pg = tilesOf(producerSet, g);
  // disjointness: Pg disjoint from Pg' for g' < g (cache prior sets)
  auto Cg = tilesOf(consumerSet, g);
  for (int64_t c : Cg)
    if (!Pg.count(c)) return emitOpError("consumer not subset of producer");
  bool allReduce   = (Cg == Pg);
  bool reduceToOne = (Cg.size() == 1);
  if (!allReduce && !reduceToOne)
    return emitOpError("reduce-to-subset unsupported");
}
```

**Verdict.** Dead simple; no Presburger link (`MLIRAnalysis`/`MLIRPresburger`
dropped). Mode gate (`|C|==1`, `C==P`) is trivial — `.size()` and `==` on sets.
But: requires *both* `groups` and a tile-id bound `T` to be known and small;
the tile bound isn't currently an input (needs a convention or extra attr); never
handles symbolic/unbounded; can blow up for large grids. This is literally
"evaluate all affine sets."

---

## Side-by-side

| | A — pure Presburger | B — pure enumeration |
|---|---|---|
| Enumerate `g` | no | yes |
| Enumerate tile ids | no | yes (needs bound `T`) |
| Handles unbounded `groups` | ✅ | ❌ |
| Disjointness | clean (2-copy + split) | trivial (set ∩) |
| Subset `C ⊆ P` | 1 call (`isSubsetOf`) | trivial |
| Mode gate `\|C\|==1` | fiddly (2nd tile copy + split) | trivial (`.size()`) |
| Presburger link dep | yes | **no** |
| Scales to large grids | ✅ | ❌ |
| Code complexity | relation plumbing | a couple of loops |

---

## Recommendation (for discussion)

- If the verifier must be **robust to symbolic/unbounded `groups`** and we accept
  the relation-plumbing cost → **Option A**, and either keep the `|C|==1` gate
  via the two-point construction *or* relax the gate to "subset only" (making
  all-reduce vs reduce-to-one informational, which loosens the policy we
  committed).
- If inter-tile `groups` are **always small bounded grids in practice** and
  simplicity wins → **Option B**, dropping the Presburger dependency entirely;
  the mode gate stays trivial. The open requirement is a **tile-id bound** input.

Any other reasons for keeping Presburger?

include/Ktdp/KtdpTypes.td
```diff
@@ -213,4 +213,58 @@ def Ktdp_RuntimeArgType : KTDP_Type<"RuntimeArg", "runtime_arg"> {
   let genVerifyDecl = 1;
 }
 
+//===----------------------------------------------------------------------===//
+// TileFutureType
+//===----------------------------------------------------------------------===//
+
+def Ktdp_TileFutureType : KTDP_Type<"TileFuture", "tile_future"> {
```
bmahjour, line 220: please see my comment [here](https://github.com/torch-spyre/ktir-mlir-frontend/pull/23#discussion_r3599398516)

fabianlim: thank you this was a great suggestion, I have updated the tile-future implementation to have the groups as an attribute of it. this removes the need now to redundently put it in the `inter-tile-produce` and `-reduce`

lib/Ktdp/KtdpTypes.cpp
```diff
@@ -151,3 +152,75 @@ Type RuntimeArgType::parse(AsmParser &parser) {
...
+// TileFutureType
+//===----------------------------------------------------------------------===//
+
+LogicalResult TileFutureType::verify(
+    function_ref<InFlightDiagnostic()> emitError, ArrayRef<Type> partialTypes,
+    IntegerSet groups) {
+  if (partialTypes.empty())
+    return emitError() << "tile_future must carry at least one partial type";
+
+  for (Type t : partialTypes) {
```
KFAFSP: Instead of this verifier, the type parameter should require `AnyRankedTensor`.

include/Ktdp/KtdpTypes.td
```diff
@@ -213,4 +213,65 @@ def Ktdp_RuntimeArgType : KTDP_Type<"RuntimeArg", "runtime_arg"> {
...
+  let summary = "Workgroup-visible handle to in-flight per-tile partial contributions";
+  let hasCustomAssemblyFormat = 1;
```
KFAFSP: nit: I would prefer it if this had a declarative assembly format, but I admit that it is impossible with the current syntax. None of the available list parsers have look-ahead and thus can't detect `groups` to end the list.

Could we chose a look-ahead free format instead?

lib/Ktdp/KtdpDialect.cpp
```diff
@@ -42,7 +42,8 @@ void KtdpDialect::initialize() {
   >();
   addTypes<
```
KFAFSP, line 43: This should use `#define GET_TYPEDEF_LIST` with the generated include.

include/Ktdp/KtdpOps.td
```diff
@@ -73,6 +73,169 @@ def KTDP_RegionTerminatorOp : Ktdp_Op<"region_terminator", [Terminator]> {
...
+
+  let arguments = (ins
+    Ktdp_TileFutureType:$future,
+    Variadic<AnyTensor>:$identity,
+    Builtin_IntegerSetAttr:$consumer_tiles_per_group,
+    OptionalAttr<Builtin_IntegerSetAttr>:$producer_dependency_per_consumer);
+  let results = (outs Variadic<AnyTensor>:$results);
+  let regions = (region SizedRegion<1>:$combiner);
+
+  let hasCustomAssemblyFormat = 1;
```
KFAFSP: I don't see why this needs a custom assembly format instead of a declarative one.

include/Ktdp/KtdpOps.td
```diff
@@ -73,6 +73,169 @@ def KTDP_RegionTerminatorOp : Ktdp_Op<"region_terminator", [Terminator]> {
...
+  let hasVerifier = 1;
+
+  let extraClassDeclaration = [{
+    /// Convenience: the partial types carried by the consumed future.
+    ::llvm::ArrayRef<::mlir::Type> getPartialTypes() {
+      return ::llvm::cast<TileFutureType>(getFuture().getType())
```
KFAFSP: This cast is redundant because `getFuture()` should already be a `TypedValue<TileFutureType>`.

include/Ktdp/KtdpOps.td
```diff
@@ -73,6 +73,169 @@ def KTDP_RegionTerminatorOp : Ktdp_Op<"region_terminator", [Terminator]> {
...
+          .getPartialTypes();
+    }
+    /// The groups set, read from the operand tile_future type.
+    ::mlir::IntegerSet getGroups() {
+      return ::llvm::cast<TileFutureType>(getFuture().getType()).getGroups();
```
KFAFSP: Same as above.

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,476 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+  IntegerSetAttr producerSet;
+  if (parser.parseKeyword("producer_tiles_per_group") || parser.parseEqual() ||
+      parser.parseAttribute(producerSet, "producer_tiles_per_group",
+                            result.attributes))
+    return failure();
+
+  if (parser.parseOptionalAttrDict(result.attributes)) return failure();
+
+  // Partial types ... -> future type.
+  SmallVector<Type> partialTypes;
```
KFAFSP: These don't do anything. Also, they are already in the future type.

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,476 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+    return failure();
+  return success();
+}
+
+void InterTileProduceOp::print(OpAsmPrinter& p) {
+  p << " producer_tiles_per_group = " << getProducerTilesPerGroupAttr();
+  p.printOptionalAttrDict((*this)->getAttrs(),
+                          /*elidedAttrs=*/{"producer_tiles_per_group"});
+  p << " : ";
+  llvm::interleaveComma(getPartialTypes(), p);
```
KFAFSP: These don't do anything.

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,476 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+  p << " -> " << getFuture().getType() << ' ';
+  p.printRegion(getBody(), /*printEntryBlockArgs=*/true,
+                /*printBlockTerminators=*/true);
+}
+
+LogicalResult InterTileProduceOp::verify() {
+  auto futureType = cast<TileFutureType>(getFuture().getType());
```
KFAFSP: Again, this should be a redundant cast.

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,476 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+  ArrayRef<Type> partials = futureType.getPartialTypes();
+
+  // Single-use invariant (§2.3): exactly one delivery op consumes the future.
+  if (!getFuture().hasOneUse() && !getFuture().use_empty())
+    return emitOpError("future result must have exactly one use (the single "
```
KFAFSP: nit: This is technically not a legal verifier according to MLIR specifications, but I understand we have a few of these around anyways.

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,476 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+
+  // Producer region: one block, single `index` group-id argument.
+  Block& block = getBody().front();
+  if (block.getNumArguments() != 1 ||
+      !block.getArgument(0).getType().isIndex())
+    return emitOpError(
+        "producer region must take a single `index` group-id argument");
+
+  // Terminator yields one value per partial role, matching the future types.
+  auto yield = cast<YieldPartialOp>(block.getTerminator());
```
KFAFSP: Accessing a nested operation must go in a region verifier (`let verifyWithRegions = 1;`).

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,476 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+
+  if (parser.resolveOperand(future, futureType, result.operands) ||
+      parser.resolveOperands(identityOperands, identityTypes,
+                             parser.getNameLoc(), result.operands))
+    return failure();
+  result.addTypes(resultTypes);
+
+  // Operand segment sizes: 1 future + N identities.
+  result.addAttribute(
+      "operandSegmentSizes",
```
KFAFSP: This operation does not implemented the `AttrSizedOperandSegments` trait, and it does not need to. This should go.

Otherwise, and only if the custom printer/parser can't be removed for some reason, do not use the hard-coded name. Use the `getOperandSegmentSizesAttrName()` provided by the trait.

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,476 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+  p.printRegion(getCombiner(), /*printEntryBlockArgs=*/true,
+                /*printBlockTerminators=*/true);
+}
+
+LogicalResult InterTileReduceOp::verify() {
+  ArrayRef<Type> partials = getPartialTypes();
+  size_t n = partials.size();
+
+  // identity: one per role, matching the partial type T_p_i.
+  if (getIdentity().size() != n)
```
KFAFSP: nit: You can replace this entire block with a `RangedTypesMatchWith` in ODS.

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,476 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+           << n << " arguments, got " << block.getNumArguments();
+  for (size_t i = 0; i < n; ++i) {
+    if (block.getArgument(i).getType() != partials[i] ||
+        block.getArgument(n + i).getType() != partials[i])
+      return emitOpError("reducer region argument pair #")
+             << i << " must both have partial type " << partials[i];
+  }
+
+  // Terminator yields N values of the partial types T_p_i.
+  auto yield = cast<YieldReducedOp>(block.getTerminator());
```
KFAFSP, line 1420: Accessing a nested operation must go in a region verifier (`let verifyWithRegions = 1;`).

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,476 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+        break;
+      }
+    }
+    if (!found)
+      return emitOpError("result #")
+             << i << " shape does not match partial #" << i
+             << " with a single unit within-group tile axis collapsed";
+  }
+
+  // Consumer-vs-producer relation (§4.1, §4.5, open question Q1).
```
KFAFSP: nit: The checks past this point are not legal in a verifier according to the MLIR specification.

include/Ktdp/KtdpOps.td
```diff
@@ -73,6 +73,169 @@ def KTDP_RegionTerminatorOp : Ktdp_Op<"region_terminator", [Terminator]> {
...
+        %s = linalg.add ins(%lhs, %rhs : tensor<1x64xf16>, tensor<1x64xf16>)
+                        outs(%lhs : tensor<1x64xf16>) -> tensor<1x64xf16>
+        ktdp.yield_reduced %s : tensor<1x64xf16>
+    }
+    ```
+  }];
+
+  let arguments = (ins
+    Ktdp_TileFutureType:$future,
+    Variadic<AnyTensor>:$identity,
```
KFAFSP: Can this really be `AnyTensor`? It seems to me that the code assumes its `AnyRankedTensor`. Same goes for the result type. This would also make some casts redundant.

README.md
```diff
@@ -18,8 +18,13 @@ The `ktdp` dialect models tile-based, data-parallel kernels targeting multi-core
 - **Data movement**: `load`, `store` (driven by access tiles)
 - **Tile identity**: `get_compute_tile_id`
 - **Runtime args**: `runtime_arg` type + `runtime_arg_extract` op
+- **Inter-tile communication**: `inter_tile_produce` + `inter_tile_reduce`
+  (all-reduce; `yield_partial` / `yield_reduced` terminators). See
+  [docs/inter-tile-communication.md](docs/inter-tile-communication.md). Broadcast
```
KFAFSP: That file doesn't exist?

lib/Ktdp/KtdpTypes.cpp
fabianlim: Fixed in e4d4e2b — switched to declarative type constraints with `AnyRankedTensor`, eliminating the manual verifier.

include/Ktdp/KtdpTypes.td
fabianlim: Fixed in 978680b — switched to look-ahead-free syntax for `tile_future` type.

lib/Ktdp/KtdpDialect.cpp
fabianlim: Fixed in 882a55c — now uses `GET_TYPEDEF_LIST` macro for type registration.

include/Ktdp/KtdpOps.td
fabianlim: Fixed in 7d1bc35 — switched to declarative `assemblyFormat`; also added the missing zero-arg `YieldPartialOp` builder required by `ensureTerminator` to unblock this.

include/Ktdp/KtdpOps.td
fabianlim: Fixed in b3ee958 — removed redundant casts since `getFuture()` already returns `TypedValue<TileFutureType>`.

include/Ktdp/KtdpOps.td
fabianlim: Fixed in b3ee958 — same redundant cast removed.

lib/Ktdp/KtdpOps.cpp
fabianlim: Fixed in d4d756b (replaced shape checks with `RangedTypesMatchWith`) and 2e185d0 (removed dead partial types from syntax).

lib/Ktdp/KtdpOps.cpp
fabianlim: Fixed in b3ee958 — removed redundant helpers/casts.

lib/Ktdp/KtdpOps.cpp
fabianlim: Fixed in b3ee958 — redundant cast removed.

lib/Ktdp/KtdpOps.cpp
fabianlim: Fixed in 947cc46 — cross-op checks moved out of the verifier.

lib/Ktdp/KtdpOps.cpp
fabianlim: Fixed in 6612651 — switched to `hasRegionVerifier` so nested-op access is in the region verifier.

lib/Ktdp/KtdpOps.cpp
fabianlim: Fixed in 9cd17ab — removed hardcoded `operandSegmentSizes`.

lib/Ktdp/KtdpOps.cpp
fabianlim: Fixed in d4d756b — replaced block with `RangedTypesMatchWith` constraint.

lib/Ktdp/KtdpOps.cpp
fabianlim: Fixed in 6612651 — moved to `hasRegionVerifier`.

lib/Ktdp/KtdpOps.cpp
fabianlim: Fixed in 947cc46 — checks moved out of the verifier.

include/Ktdp/KtdpOps.td
fabianlim: Fixed in e4d4e2b — `AnyTensor` replaced with `AnyRankedTensor`, which makes the downstream casts redundant.

README.md
fabianlim: this file will exist once #23 is merged

include/Ktdp/KtdpOps.td
KFAFSP: You addressed the declarative format for `ktdp.inter_tile_produce`, which is great, but not for this `ktdp.inter_tile_reduce` op.

include/Ktdp/KtdpTypes.td
KFAFSP: The point of this change was to be able to use a declarative assembly format. I agree that the original syntax looks better, so if that is not possible, there's no point to the change.

To enable this, you also have to change `IntegerSet` to `IntegerSetAttr`, which is what you're using for printing & parsing anyways. 

`AffineMap` also supports a bare `mlir::FieldParser` that comes without the `affine_map` part around it. Since nobody uses it, nobody bothered to add this for `IntegerSet` it seems. Thus you could also provide template specializations for `mlir::FieldParser<mlir::IntegerSet>` and `operator<<` in the `.cpp` file and use the unqualified set syntax.

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,319 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+  // collapsed) is expressed declaratively via RangedTypesMatchWith in the op
+  // definition (see KtdpOps.td), which calls TileFutureType::getReducedPartialTypes().
+
+  // Local structural checks on the consumer set attribute.
+  IntegerSet consumerSet = getConsumerTilesPerGroup().getValue();
+  if (consumerSet.getNumSymbols() != 1)
+    return emitOpError("`consumer_tiles_per_group` must have exactly one "
+                       "symbol (the group index g)");
+
+  // TODO: move producer/consumer set relationship checks to a dedicated pass.
```
KFAFSP: I appreciate you moving this out of the verifier. Having these kinds of checks in here breaks the dialect conversion infrastructure, so even though it appears convenient, it makes the dialect harder to use.

However, I don't think there should be a TODO here. Instead of [this commit that removes everything](https://github.com/torch-spyre/ktir-mlir-frontend/commit/947cc467ffa41b434b427b4e846af603319a2f7e), why don't you put it into a method for now? We can always call that from the pass or move it later.

KFAFSP:
Please note that I'm not allowed to resolve review conversations without write access (apparently a GitHub thing), but I'm happy about all the other changes!

include/Ktdp/KtdpTypes.td
fabianlim: > The point of this change was to be able to use a declarative assembly format. I agree that the original syntax looks better, so if that is not possible, there's no point to the change.

Im abit confused by this. Was the choice to add paranthesis, to address. your previous comment to have. a look-ahead free format,  incorrect? Im not clear what "To enable this" means.. what does "this" in this sentence refer to?

fabianlim: On the other hand, if I were to interpret your comment as if you are ok with the paranthesis, then if I take your suggestion to change to IntegerSetAttr, this is addressed in e6346a5, but if I understood wrongly I can revert it

include/Ktdp/KtdpOps.td
fabianlim: Addressed in 59aa726 — `InterTileReduceOp` now uses a declarative `assemblyFormat`, removing the hand-written `parse`/`print` (~90 lines). One minor syntax change: multi-valued identity prints as `%a, %b : T1, T2` instead of `%a : T1, %b : T2` since that is what the declarative variadic format produces (single-valued case is unchanged).

lib/Ktdp/KtdpOps.cpp
fabianlim: I can move this to a function if you like, but before that wanted to address https://github.com/torch-spyre/ktir-mlir-frontend/pull/25#discussion_r3620589205

include/Ktdp/KtdpTypes.td
KFAFSP: Sorry for the confusion, let me explain it.

I think that we should always use declarative dialect definitions whenever possible. That includes invariants (leading to verifiers) as well as assembly formats. The point is to not write any C++ code, be free from maintenance issues, have TableGen double-check your work, make it easier to read, and to assist in bumping MLIR versions.

For custom printing & parsing, there are certain formats we can't achieve with the declarative form alone. In particular, the builtin parsers can't deal with arbitrary look ahead. You need delimited groups of tokens. My stance is that, if required, such parts of the format should be extracted into a [custom directive](https://mlir.llvm.org/docs/DefiningDialects/Operations/#custom-directives) (some of those already exist upstream) and used in the declarative format instead.

You've done exactly what I would have liked to see with this change, which is to eliminate the look ahead and allowing us to use a built-in printer & parser. But you did not do that final step. Your type definition should read:

```td
let assemblyFormat = "`<` `(` $partialTypes `)` `,` `groups` `=` $groups `>`";
// ...
let parameters = (ins
    ArrayRefParameter<"RankedTensorType">:$partialTypes,
    "IntegerSetAttr":$groups
  );
```

and then you don't need to have `IntegerSetAttrParam` or `PartialTypesParam` at all. This was enabled by changing to `IntegerSetAttr` instead of `IntegerAttr` for `$groups`.

---

The second part of my comment was about the fact that `IntegerSetAttr` works as a type parameter because it is derived from `Attribute` and therefore has an implicit `mlir::FieldParser`. `IntegerSet` does not, while `AffineMap` does. This is somewhat of an inconsistency that can easily be remedied in our code _if_ we wanted to use _exactly_ that type. The difference being in the resulting format though:

Instead of `affine_set<(i)[g] : (i - 4*g >= 0, -i + 4*g + 3 >= 0)>` as produced by the `IntegerSetAttr`, the `AsmParser::parseIntegerSet` function we'd use in the `mlir::FieldParser` omits the attribute part and expects `(i)[g] : (i - 4*g >= 0, -i + 4*g + 3 >= 0)`. For upstream users of `AffineMap` parameters, this is what they intend to have.

I do not have any strong preferences towards either of those, except that the first option (that you went with) requires less code on our side.

fabianlim: thank you for the clarification. I have addressed the first part in [b4f78f1](https://github.com/torch-spyre/ktir-mlir-frontend/pull/25/commits/b4f78f1c4f131a9cb6a638166fd0498f44b6679b). 

The second part seems to be more complicated and it will impact (as you said) other exiting ops like `construct_memory_view` and `construct_access_tile` etc. Thus im a little concerned if we made that change, then we will affect other repos like ktir-cpu, triton, etc.. For this I created an issue https://github.com/torch-spyre/ktir-mlir-frontend/issues/34 and maybe we can discuss as a follow up

fabianlim:
THanks @KFAFSP for the approval, but just wanted to discuss the need to have the dimension reduction in the `inter_tile_reduce`. For example in `test/Ktdp/inter-tile-reduce-roundtrip.mlir` ,we see that in the `inter_tile_reduce` region is typed `<(tensor<128x1x1x64xf16>), groups = #mg_groups> -> tensor<128x1x64xf16>`, see below

```mlir
func.func @reduce_multi_group(%partial: tensor<128x1x1x64xf16>,
                              %add_id: tensor<128x1x1x64xf16>) -> tensor<128x1x64xf16> {
  %f = ktdp.inter_tile_produce
      producer_tiles_per_group = #mg_tiles
      -> <(tensor<128x1x1x64xf16>), groups = #mg_groups>
  {
    ^bb0(%gid: index):
      ktdp.yield_partial %partial : tensor<128x1x1x64xf16>
  }
  %r = ktdp.inter_tile_reduce(%f)
      consumer_tiles_per_group = #mg_tiles,
      identity(%add_id : tensor<128x1x1x64xf16>)
      : <(tensor<128x1x1x64xf16>), groups = #mg_groups> -> tensor<128x1x64xf16>
  {
    ^bb0(%lhs: tensor<128x1x1x64xf16>, %rhs: tensor<128x1x1x64xf16>):
      %s = linalg.add ins(%lhs, %rhs : tensor<128x1x1x64xf16>, tensor<128x1x1x64xf16>)
                      outs(%lhs : tensor<128x1x1x64xf16>) -> tensor<128x1x1x64xf16>
      ktdp.yield_reduced %s : tensor<128x1x1x64xf16>
  }
  return %r : tensor<128x1x64xf16>
}
```

Its not obvious, but there was a `1x` dimension being dropped. This was from our earlier [discussion](https://github.com/torch-spyre/ktir-mlir-frontend/pull/22#discussion_r3263307521) in the design proposal with the desire to follow `linalg.reduce` semantics. But @KFAFSP pointed out to me one key difference with `inter_tile_reduce` vs `linalg.reduce`, is that the "whole" shape of the input does not show up in the MLIR. In the above example, the reduce is over the affine set `mg_tiles` tiles which is **implicit**. As such I feel not making the rank-reduction explicit is easier for the progammer and also makes the impelemntation of this new op simpler. 

@vswagath1989 can we consider not to have the rank-reduction, or have it as an optional feature (personally i prefer not to have it as the way we decide which dimension in `getReducedPartialTypes` is over complicated.). We dont have to decide rightaway, and we could stage this, i could implement this first without rank-reduction, and we can have an issue to deliberate it

KFAFSP:
I like having less implicit things going on, also because it allows us to forgo the rank expansion before this op and thus have simpler IR. I would also like to point out that `tensor<f16>` is a gracefully handled edge case for upstream MLIR, so the 1 is not necessary to protect us from these.

My argument was: the equivalent `linalg.reduce` would _not_ remove the 1 dimension, because that is not the one we are reducing over. That one is implicit in this IR.

vswagath1989:
@fabianlim @fabianlim - I don't see a need for rank reduction (reduction in dimensionality of the tensor) either. Let us go with equal number of dimensions for input and output. thanks!

fabianlim:
@vswagath1989 @KFAFSP I pushed one more commit 507a545  to remove the implicit rank reduction

include/Ktdp/KtdpOps.td
```diff
@@ -175,11 +175,7 @@ def KTDP_InterTileProduceOp : Ktdp_Op<"inter_tile_produce",
 
 def KTDP_InterTileReduceOp : Ktdp_Op<"inter_tile_reduce",
     [SingleBlockImplicitTerminator<"::mlir::ktdp::YieldReducedOp">,
-     RecursiveMemoryEffects,
-     RangedTypesMatchWith<
```
KFAFSP: Now you don't have the result type check at all anymore. That's not good, please add it back.

include/Ktdp/KtdpOps.td
```diff
@@ -73,6 +73,178 @@ def KTDP_RegionTerminatorOp : Ktdp_Op<"region_terminator", [Terminator]> {
...
+  let results = (outs Variadic<AnyRankedTensor>:$results);
+  let regions = (region SizedRegion<1>:$combiner);
+
+  let assemblyFormat = [{
+    `(` $future `)` `consumer_tiles_per_group` `=` $consumer_tiles_per_group `,`
+    (`producer_dependency_per_consumer` `=` $producer_dependency_per_consumer^ `,`)?
+    `identity` `(` ($identity^ `:` type($identity))? `)`
+    attr-dict `:` type($future) `->` type($results) $combiner
+  }];
+  let hasRegionVerifier = 1;
```
KFAFSP, line 243: Actually, I somehow overlooked that you replaced the `::verify` method with `::verifyRegions` entirely. That is not quite appropriate. You should have both of them, with the former handling op-local checks and the latter doing checks on children that you may assume to be locally verified (that's the whole point of `verifyRegions`).

include/Ktdp/KtdpOps.td
fabianlim: Addressed in d98d1a9 — added a result type check in `InterTileReduceOp::verify()` that enforces each result type equals the corresponding partial type exactly (same type, no rank reduction). If this is not the right thing to do please let me know

include/Ktdp/KtdpOps.td
fabianlim: we added a `hasVerifier = 1` and split the checks into the two verification methods

fabianlim: the round-trip changed because it turns out that without the verify previously the check was not firing. So it was a good catch!

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,241 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+  return out;
+}
+
+}  // namespace
+
+// Syntax:
+
+LogicalResult InterTileProduceOp::verify() {
+  // Single-use invariant (§2.3): exactly one delivery op consumes the future.
+  if (!getFuture().hasOneUse() && !getFuture().use_empty())
```
KFAFSP: Now you've added this illegal check back into the verifier.

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,241 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+
+  return success();
+}
+
+LogicalResult InterTileProduceOp::verifyRegions() {
+  TileFutureType futureType = getFuture().getType();
+  ArrayRef<RankedTensorType> partials = futureType.getPartialTypes();
+
+  // Producer region: one block, single `index` group-id argument.
+  Block& block = getBody().front();
```
KFAFSP: This check _is_ allowed in the normal verifier, since it doesn't actually visit operations within the region.

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,241 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+             << " does not match future partial type " << partials[i];
+
+  return success();
+}
+
+//===----------------------------------------------------------------------===//
+// InterTileReduceOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult InterTileReduceOp::verify() {
```
KFAFSP, line 1390: Now you've made the verifier non-declarative again. The first two checks should be handled by traits. For example, this should work:

```
RangedTypesMatchWith<"", "future", "results", "llvm::cast<TileFutureType>($_self).getPartialTypes()">,
TypesMatchWith<"", "results", "identity", "$_self">,
```

lib/Ktdp/KtdpOps.cpp
```diff
@@ -1219,6 +1223,241 @@ LogicalResult RuntimeArgExtractOp::verify() {
...
+
+  return success();
+}
+
+LogicalResult InterTileReduceOp::verifyRegions() {
+  ArrayRef<RankedTensorType> partials = getPartialTypes();
+  size_t n = partials.size();
+
+  // Reducer region: 2N args (lhs_1..lhs_N, rhs_1..rhs_N) each of type T_p_i.
+  Block& block = getCombiner().front();
```
KFAFSP: See above, this can be part of `::verify`.

lib/Ktdp/KtdpOps.cpp
fabianlim: i thuink this one slipped through, it was still in the previous commit. I will remove it. One of the tests was removed also

lib/Ktdp/KtdpOps.cpp
fabianlim: moved to to verify but had to duplicate `Block& block = getCombiner().front()`. A little bit confused why the combiner is not considered belonging to the region

lib/Ktdp/KtdpOps.cpp
fabianlim: updated the tablegen to make this declarative

lib/Ktdp/KtdpOps.cpp
fabianlim: yes understand, moved this to verify, but had to duplicate `Block& block = getBody().front();`

lib/Ktdp/KtdpOps.cpp
KFAFSP: The combiner is the region. The point of `verifyRegions` is that it runs after the children have been verified. The region and its blocks themselves belong to the op, so you can verify them in `::verify`. You're just not allowed to visit the ops within them.

KFAFSP:
I think we should leave the conversion pass stub for another PR so we can merge this one without needing to complicate the review. What do you think?

fabianlim:
@KFAFSP 
> I think we should leave the conversion pass stub for another PR so we can merge this one without needing to complicate the review. What do you think?

sorry that was a miss push.. fixing it. I was trying to push it to my fork for you to take a look on the issue

PR #26, opened 2026-06-29
Issue 24 llvm releases

### LINKED ISSUE #24: Publish prebuilt LLVM as GitHub Releases instead of Actions artifacts
Follow-up to #14, which set up the stable LLVM artifact CI and the `setup_mlir.py` developer flow. This proposes evolving that artifact mechanism into GitHub Releases. cc: @lchu6 

## TL;DR

Serving the pinned LLVM as a **Release asset** (public, no token, no expiry)
instead of an **Actions artifact** drops the `GIT_PAT` requirement and the
scheduled retention-refresh job. Open concerns before doing it:

- **Storage isn't auto-cleaned.** Release assets never expire and nothing
  prunes them (today's flow both expires at 90d *and* deletes the old
  artifact). Need a retention policy. Mitigated by: builds only happen on a
  hash bump, so growth is slow (~700 MB each).
- **Releases imply a code version.** These are a build-input cache, not
  frontend releases — they'd add synthetic `llvm-<hash>` tags to the repo.
  Use `prerelease: true`, or a separate artifacts repo.
- **Migration:** keep the artifact path as a fallback until releases are
  proven, then drop the token requirement.

## Problem

`scripts/setup_mlir.py` downloads the pinned LLVM as a GitHub **Actions
artifact** via the API. Two costs:

1. **Needs a token** — the artifact zip endpoint requires `Authorization`, so
   `setup_mlir.py` fails without `GIT_PAT` / `GITHUB_TOKEN`. Friction for new
   contributors, secret-less CI, forks, offline.
2. **Expires (90d)** — a scheduled CI job exists *only* to re-upload and reset
   the retention clock (`docs/ci.md`, "Scheduled refresh").

## Proposal

Publish each LLVM build as a **Release asset** tagged `llvm-<hash>`. For this
public repo the asset is downloadable with no token, never expires, and the
hash tag maps 1:1 to a pin — so the PAT requirement and the refresh job both go
away.

## Changes

- **`llvm-build.yml`:** it already builds only on a hash bump (push to `main`
  path-filtered to `cmake/llvm-hash.txt`, plus refresh/PR/dispatch). On that
  build, also `gh release create llvm-<hash>` + upload the tarball; gate the
  publish on the `main`-push so PRs build-test but don't publish.
- **`setup_mlir.py`:** add a release-URL fetch path (no auth header), keeping
  the artifact path as a fallback during migration.
- **Retire** the scheduled refresh job (its only job is resetting the artifact
  clock; assets don't expire).
- **`docs/ci.md`:** update the flow table + refresh section.

## Concern: storage is never auto-cleaned

Today storage is bounded twice — 90d expiry *and* the "Delete old artifact"
step. Release assets have neither, so each hash bump leaves a permanent
~700 MB asset. Pick a policy:

1. **Keep last N** — prune older releases on bump (mirrors today's delete step).
2. **Keep everything** — slow growth (builds are rare); permanence is arguably
   a feature. Likely fine on a paid org.
3. **Hybrid** — keep tags, attach assets only to the last N.

Note: for public repos, release-asset downloads don't count against the Actions
storage/transfer quota, so billing may improve even as raw bytes grow.

## Concern: tying LLVM to "releases"

A Release needs a git tag, created by `gh release create llvm-<hash> --target
main`. These appear on the repo's Releases page but are a build-input cache, not
software releases. Either mark them `prerelease: true` (keeps them off "Latest",
synthetic `llvm-*` tags clutter `git tag`), or host them in a separate
`torch-spyre/llvm-artifacts` repo (cleaner, one more repo). If real frontend
releases ever ship, the separate repo is better.

## Pointers

- `scripts/setup_mlir.py` — `query_artifact_id`, `download_and_cache`, token in `main()`.
- `docs/ci.md` — artifact flow table + "Scheduled refresh".
- Downstream: Triton spyre build calls `setup_mlir.py` from `setup.py`
  (LLVM pin in `cmake/llvm-hash-spyre.txt`).

danielezer:
Publish LLVM builds as GitHub releases

Closes #24.

Previously LLVM builds were stored as GitHub Actions artifacts, which expire and are awkward to fetch outside CI and require an API key, even though the repo is public. This PR publishes each build as a GitHub release instead, keyed by the LLVM hash.

- llvm-build.yml: build published as a release; skips rebuild/upload when a release for the hash already exists.
- setup_mlir.py: fetches the prebuilt LLVM from the release (falls back to the legacy Actions artifact).
- docs/ci.md: documents both flows — legacy artifact and new release — plus the manual workflow_dispatch job to publish a release on demand.


fabianlim (commented):
# Review: PR #26 — Issue 24 LLVM Releases

## Summary of changes

The PR migrates LLVM build distribution from **GitHub Actions artifacts** (90-day
expiry, token-gated) to **GitHub Release assets** (permanent, public, no auth).
Three files are touched: `llvm-build.yml`, `scripts/setup_mlir.py`, `docs/ci.md`.

### `llvm-build.yml`
- Removes the bi-monthly `schedule` trigger and the entire `refresh` job
  (no more 90-day clock to fight).
- Adds a `create-release` job that runs before the 3-platform build matrix,
  creating the `llvm-<short-hash>` prerelease once (idempotent). This is the
  proposed fix for the race condition (multiple parallel jobs all trying to
  `gh release create` at once).
- Each `build` matrix job now calls `gh release upload --clobber` instead of
  `actions/upload-artifact`.
- Replaces the `refresh` job with a `prune` job that deletes old `llvm-*`
  releases beyond `LLVM_RELEASE_KEEP=10`, serialised with
  `concurrency: llvm-prune`.
- The `schedule` trigger is dropped entirely; PR builds skip `create-release`
  and `publish release asset` steps (dry-run mode).

### `scripts/setup_mlir.py`
- Adds `release_asset_available()` (HEAD request, no auth) and
  `download_release_and_cache()` (streams tarball direct from releases URL).
- Extracts `_stream_download` and `_extract_tar_to_cache` as shared helpers
  (deduplication).
- Resolution order is now:
  1. `--wheel`
  2. Cache hit
  3. **GitHub Release asset** (no token) ← new primary network path
  4. Actions artifact (token-gated) ← legacy fallback
  5. Error if neither available and no token

### `docs/ci.md`
- Updated throughout to reflect the new release-asset model.
- Removes the scheduled-refresh description.
- Clarifies that the legacy Actions-artifact fallback is token-gated and only
  used for hashes built before releases were adopted.

---

## What needs to change in the PR

The PR rips out the old artifact path in one shot. We want a staged migration
where the release path is added *alongside* the existing artifact path, proven
to work, and only then the old path is removed. Concretely:

1. **Do not remove `actions/upload-artifact`.** Keep it in the `build` job
   alongside the new `gh release upload`, under the same `exists != 'true'`
   guard, with a comment marking it for deletion once the release path is
   confirmed. Both run during migration so there is always a working source.

2. **Do not remove the `schedule` cron / `refresh` job.** The cron keeps the
   Actions artifacts alive (90-day clock) while they are still the fallback.
   Remove it only after the release path is confirmed and the artifact path is
   dropped.

3. **`LLVM_RELEASE_KEEP` / `prune` job.** The `prune` job lists `llvm-*`
   releases, sorts newest-first, and deletes everything past index
   `LLVM_RELEASE_KEEP` (=10). It exists purely to bound storage now that
   releases never expire. Per the plan, we are *not* doing automated pruning
   yet (few LLVM versions expected). So: **drop the `prune` job and the
   `LLVM_RELEASE_KEEP` env var for now**, and replace with a one-line note in
   `docs/ci.md` giving the manual delete command
   (`gh release delete <tag> --cleanup-tag --yes`).

4. **`build_release_url` parse is fragile.** It derives the tag via
   `artifact_name.split("-")[1]`, assuming the short hash is always the second
   dash-separated field. Make this explicit (construct the tag from the known
   short-hash input rather than re-parsing the artifact name).

5. **`docs/ci.md`** — during Stage 1, document the new release flow in an
   additional paragraph and state that migration is in progress (both the
   release path and the legacy artifact/cron path are live). Keep the existing
   artifact/cron docs in place. Only in Stage 2, once the old flow is
   deprecated, do we mark that paragraph as deprecated / remove it.

---

## Target flow (Stage 1 end-state, both paths live)

This is the flow we want once Stage 1 lands — release path added *alongside* the
artifact path, so both produce/resolve in parallel during migration.

### `llvm-build.yml`

- **`get-hash` job**: read the pinned hash from `cmake/llvm-hash.txt` (or the
  dispatch override); compute `short-hash` and `release-tag = llvm-<short>`.

- **`create-release` job** (push-to-main / dispatch only; skipped on PR):
  - `gh release view <tag>` → if absent, `gh release create <tag> --prerelease`.
  - Idempotent: if the release already exists, no-op.

- **`build` job** (matrix: 3 platforms; runs on push / dispatch / PR):
  - **Check existing release asset** → `exists=true` if
    `llvm-<short>-<os>-<arch>.tar.gz` is already attached to `<tag>`.
  - **Check existing artifact** (legacy) → `exists=true` if the Actions artifact
    of the same name already exists. (Either check setting `exists=true` is
    enough to skip.)
  - **Build LLVM** only if `exists != true`.
  - **Upload to release**: `gh release upload <tag> ...tar.gz --clobber` — only
    when `create-release` ran (i.e. not a PR).
  - **Upload to artifact** (legacy, `actions/upload-artifact`): keep, under the
    same `exists != true` guard, marked `# DEPRECATED (issue #24)`.

- **`schedule` cron + `refresh` job**: keep (deprecated, keeps artifacts within
  the 90-day window while they remain a fallback).

- **No `prune` job / no `LLVM_RELEASE_KEEP`** — manual delete documented in
  `ci.md` instead.

### `ci.yml` → `setup_mlir.py`

- **Resolve from release first**: HEAD-check the public `llvm-<short>` release
  asset and download it — **no `GIT_PAT`** needed.
- **Fall back to artifact**: only if the release is absent, query the Actions
  artifact API, which **requires** `GIT_PAT` / `GITHUB_TOKEN`.
- **`Set up MLIR` step**: **comment out** `GITHUB_TOKEN` in the `env:` so CI
  exercises the release path with no token (this is the thing we want to test).
  Leave it commented as documentation + ready fallback:
  `# GITHUB_TOKEN: ... # uncomment to fall back to the legacy Actions artifact`.

---

## Recommended merge sequence (staged)

Releases must exist before anything is told to read from them, so the build
workflow (which *produces* releases) goes first.

### Stage 1 — Land the release-producing build, keep the artifact fallback

Merge `setup_mlir.py` (release path + artifact fallback) and the
`llvm-build.yml` changes, but **with the artifact upload and cron kept** (items
1–3 above). After merge:

1. Trigger a real build (`workflow_dispatch` against a hash with no existing
   release, or a `cmake/llvm-hash.txt` bump) and confirm the
   `create-release → build → publish` chain produces a populated
   `llvm-<short-hash>` release with all three platform assets.
2. Confirm the artifact upload still ran too (fallback intact).
3. Run `scripts/setup_mlir.py` locally **without `GIT_PAT`** and confirm it
   resolves via the release asset.

Note on what gates `build`: the `build` job runs on push-to-main (hash bump),
`workflow_dispatch`, **and** PRs touching `llvm-build.yml`. Its `if` is
`!cancelled() && llvm-hash != '' && create-release.result != 'failure'` — it is
*not* gated on "a release exists." On a PR, `create-release` is *skipped* (not
failed), so `build` still runs to build + package and validate the workflow; it
just skips the `create-release`/`publish` steps. So a real release is produced
only on push-to-main or dispatch, but the build itself runs in all three cases.

### Stage 2 — Cut the consumers over, then remove the legacy path

1. Open a PR on [`ktir-cpu` CI](https://github.com/torch-spyre/ktir-cpu/blob/main/.github/workflows/ci.yml#L32-L41)
   removing `GIT_PAT` from the `setup_mlir.py` call; confirm its CI passes with
   no token.
2. Once that is green: **deprecate** (do not delete) the legacy artifact path —
   comment out the `actions/upload-artifact` step, the `schedule` trigger, and
   the `refresh` job, with a `# DEPRECATED (issue #24): ...` header explaining
   it can be re-enabled if the release path regresses. (Alternative if we
   prefer keeping them runnable: guard them behind an `if:` on a workflow input
   / repo variable defaulting to off, rather than commenting out.)
3. Update `docs/ci.md`: mark the legacy-flow paragraph as deprecated and make
   the release flow the documented default; include the manual delete note from
   item 3.

danielezer:
Thanks for the detailed review @fabianlim, I made the changes you requested.

docs/ci.md
```diff
@@ -35,53 +35,60 @@ switches all three flows to the custom artifact path automatically.
...
-the artifact expires.  A full rebuild is unnecessary — the content is
-identical.
-
-1. Reads the hash from `cmake/llvm-hash.txt`
-2. Checks GitHub Actions artifacts — existing artifact found
-3. Downloads the artifact zip via the GitHub API
-4. Re-uploads it — a new artifact entry is created with a fresh 90-day clock
-5. Deletes the old artifact to prevent stale duplicates accumulating in GitHub Actions
-6. Flow 1 is **not** triggered (artifact content unchanged)
+```bash
```
fabianlim, line 108: I think we shouldnt overwrite the artifacts flow for now, since we are going to depracate later. I would structure the document as having two flows. You can make the artifact flows are scheduled for depracation and promote the release flow. Just make it clear there are two existing mechanisms now

docs/ci.md
```diff
@@ -101,7 +108,7 @@ gh workflow run llvm-build.yml -f llvm-hash=<full-40-char-sha>
 
 **What it does:** runs the same build + test pipeline as Flow 1 but sources
 MLIR from the latest `mlir_wheel` on the eudsl index instead of the pinned
-artifact.
+release.
```
fabianlim: the mlir_wheel theoretically is not a release, its a nightly build. I think just keep the old wording

docs/ci.md
fabianlim: If it makes it easier organizationally, you can have two subsections


### Scheduled Refresh

#### Release Flow

There is none, the release artifacts do not expire, and there is no need for refreshing.

#### Artifacts Flow [Marked for Deprecation]

...

docs/ci.md
```diff
@@ -101,7 +108,7 @@ gh workflow run llvm-build.yml -f llvm-hash=<full-40-char-sha>
 
 **What it does:** runs the same build + test pipeline as Flow 1 but sources
 MLIR from the latest `mlir_wheel` on the eudsl index instead of the pinned
-artifact.
```
fabianlim, line 104: Same as above, the release flow also has a manual trigger flow, that is to create a release when we do not bump the llvm-cache.txt. So that has to be documented as well.
1. Ensure that release does not exist
2. ...

docs/ci.md
```diff
@@ -35,53 +35,60 @@ switches all three flows to the custom artifact path automatically.
...
+  while it remains a fallback. Removed once the release path is confirmed.
 
-**What it does** depends on whether a valid artifact already exists:
+**What it does** depends on whether the release asset already exists:
 
-### Hash bump (artifact does not exist)
+### Hash bump (release asset does not exist)
 
 1. Reads the new hash from `cmake/llvm-hash.txt`
-2. Checks GitHub Actions artifacts — none found for this hash
```
fabianlim, line 62: Same as below, just document two flows separately for now marking artifacts for depracation

CLAUDE.md
```diff
@@ -0,0 +1,101 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## What this is
+
+KTIR (Kernel Tile IR) frontend: an out-of-tree MLIR dialect named `ktdp` (Kernel
+Tile Data-Parallel) that models tile-based, data-parallel kernels targeting
+multi-core accelerators (Spyre). The dialect supplies **only** memory-view,
+access-tile, and data-movement constructs; compute and control flow are
```
fabianlim: If we want to commit this file, it should be done in a separate PR

.gitignore
```diff
@@ -1,3 +1,5 @@
 build
 *.egg-info
-*.pyc
\ No newline at end of file
+*.pyc
```
fabianlim: lets not pollute the .gitignore with code editor specific items

.github/workflows/llvm-build.yml
```diff
@@ -61,33 +102,59 @@ jobs:
...
+          ARTIFACT_NAME="llvm-${SHORT}-${{ matrix.config.os }}-${{ matrix.config.arch }}"
           echo "artifact-name=${ARTIFACT_NAME}" >> $GITHUB_OUTPUT
-          ARTIFACT_ID=$(curl -sf \
-            -H "Authorization: Bearer $GH_TOKEN" \
-            "https://api.github.com/repos/${{ github.repository }}/actions/artifacts?name=${ARTIFACT_NAME}&per_page=1" \
-            | jq -r '.artifacts[0].id // empty')
-          if [ -n "$ARTIFACT_ID" ]; then
-            echo "exists=true" >> $GITHUB_OUTPUT
-            echo "::notice::Artifact ${ARTIFACT_NAME} already exists (id=${ARTIFACT_ID}); skipping build"
+
```
fabianlim: as mentioned in the review. lets not mix logic like this. Lets have two separate jobs, one for release and one for artifacts

.github/workflows/llvm-build.yml
```diff
@@ -61,33 +102,59 @@ jobs:
...
+            echo "::notice::Release asset ${ARTIFACT_NAME}.tar.gz already exists"
           else
-            echo "exists=false" >> $GITHUB_OUTPUT
+            RELEASE_EXISTS=false
           fi
+          echo "release-exists=${RELEASE_EXISTS}" >> $GITHUB_OUTPUT
 
-      - uses: actions/checkout@v4
-        if: steps.check.outputs.exists != 'true'
+          # DEPRECATED (issue #24): legacy Actions-artifact existence check.
```
fabianlim: for the control flows, is best to use `check.outputs` its more clean

danielezer: I'm not sure this is valid syntax (according to google)

.github/workflows/llvm-build.yml
danielezer: A completely different job for each workflow would mean two llvm builds x3 archs. Each such build takes about 30 mins, and doing this twice seems a bit wasteful, since they build the same artifact. 

The only thing that differs between the two workflows with respect to the build is the if exists check, and the gating in the if statements for following steps.

Are you sure you want me to create a different job?

docs/ci.md
```diff
@@ -35,64 +35,115 @@ switches all three flows to the custom artifact path automatically.
...
+after an accidental delete, or a one-off hash). A single dispatch builds once
+and feeds both flows, each upload step skipping if its own target already
+exists — so check first. The check differs per flow:
+
+#### Release Flow
 
 ```bash
-# Rebuild for the current pinned hash (e.g. after accidental artifact deletion):
-gh workflow run llvm-build.yml
+gh release view llvm-<short-hash>            # public, no token
```
fabianlim, line 109: i wasnt sure if this will work, i thought that the flow must be triggered from the workflow, does it trigger from creating a release?

.github/workflows/llvm-build.yml
```diff
@@ -29,27 +36,61 @@ jobs:
     outputs:
       llvm-hash:       ${{ steps.read.outputs.llvm-hash }}
       llvm-short-hash: ${{ steps.read.outputs.llvm-short-hash }}
+      release-tag:     ${{ steps.read.outputs.release-tag }}
```
fabianlim, line 39: is this needed? i wasnt clear what we are going to use for the release-tag, if im not wrong, even if llvm-hash is an official release tag we will be able to fetch the llvm repo from the upstream. So if that is the case, do we need another tag?

.github/workflows/llvm-build.yml
```diff
@@ -61,33 +102,52 @@ jobs:
...
+            echo "artifact-exists=true" >> $GITHUB_OUTPUT
+            echo "::notice::Legacy artifact ${ARTIFACT_NAME} already exists"
+          else
+            echo "artifact-exists=false" >> $GITHUB_OUTPUT
+          fi
 
       - name: Set artifact name
-        if: steps.check.outputs.exists != 'true'
-        run: echo "LLVM_INSTALL_DIR=${{ steps.check.outputs.artifact-name }}" >> $GITHUB_ENV
+        if: ${{ (steps.release.outputs.release-exists != 'true' || steps.artifact.outputs.artifact-exists != 'true') && success() }}
```
fabianlim: this is fine, but if combine the two flags to a single flag, the depracation will be easier later, but minor

fabianlim: its easier to rationalize a single `steps.check.outputs.exists != 'true'` then a long check

danielezer: Sure, changed.

.github/workflows/llvm-build.yml
danielezer: This is the name of the release we upload to github releases, the variable is just for convenience

docs/ci.md
```diff
@@ -35,64 +35,115 @@ switches all three flows to the custom artifact path automatically.
...
+  once the Release Flow is confirmed and all consumers have cut over.
+
+Both mechanisms exist on purpose right now; `setup_mlir.py` resolves the release
+asset first and falls back to the artifact.
+
 **Triggers:**
 - Push to `main` that changes `cmake/llvm-hash.txt` (hash bump)
-- Scheduled: 1st of every other month at 02:00 UTC (`0 2 1 */2 *`)
 - Manual: `workflow_dispatch` with an optional hash override
+- Pull request that touches `llvm-build.yml` (builds + packages to validate the
```
fabianlim, line 70: this is already mentioned here that we have a `workflow_dispatch` flow for manual building of llvm, however the steps below are mostly manual steps. This is fine, however, most maintainers would prefer to build on the CI rather than their local machine, so would be good also to highly `workflow_dispatch` flow for the manaul case (flow 2)

danielezer: added notes that point to workflow 3 from the manual invocation info

fabianlim:
@danielezer can you fix the DCO

fabianlim (approved):
went through a few iterations, and we have a demo on @danielezer 's fork https://github.com/danielezer/ktir-mlir-frontend/pull/1
- [releases](https://github.com/danielezer/ktir-mlir-frontend/releases/tag/llvm-e9846648) 
- [github actions](https://github.com/danielezer/ktir-mlir-frontend/actions/runs/28573273034)

fabianlim:
@lchu6 @kiszk if you have time can you take a look at this PR. this will enable llvm build products in the release

fabianlim:
Added a commit to enable `LLVM_ENABLE_ASSERTIONS` for all builds (previously it was only set for nightly builds, leaving it inert for stable tags like `llvmorg-22.1.3`). This is needed for the `dataflow-scheduler` CI — consumers that set `LLVM_ENABLE_ASSERTIONS=ON` in their own build were getting no-op assertions and no `LLVM_DEBUG()` output because the linked MLIR code was compiled with `-DNDEBUG`.

Note: this will cause a diff between the new release assets and the existing Actions artifacts (the legacy artifacts were built without assertions). The Actions artifact path is being retired anyway once the release path is confirmed, so this divergence is intentional and temporary.

PR #27, opened 2026-06-30
Implemented a subproject build system for KTIR.

KFAFSP:
This patch brings KTIR's build system in line with the new dataflow scheduler family of repositories, which currently need to use a fork of `ktir-mlir-frontend` to support building it as a subproject and bundling the Python bindings.

In future, I'd also like to unify the project file structure and testing procedure. However, this PR is not about such optional changes, and only about the minimum necessary to host the new submodule-based project hierarchy.

-----

This patch changes the build system of KTIR to enable:
  - CMake packaging (`find_package`)
  - subproject builds (`add_subdirectory`)
  - linking against any MLIR (STATIC, SHARED, DSO)

This is implemented using a build system implementation that is shared between `dataflow-scheduler`, `dataflow-scheduler-mlir-dialects` and this repository. It includes new top-level targets and the mechanisms required to make installation work.

Linking against MLIR was overhauled such that this project (and the others) can be built against any distro of MLIR, whether it was linked statically, with shared libraries (developer mode), or as a single DSO (previously the only build mode). This change removes the MLIR DSO from the installation instructions (if used), instead setting RPATH to the external source.

Full list of changes:
  - Added a new build system (`AddKTIR*.cmake`).
  - Added CMake installation and packaging (`cmake/*`).
  - Added `python/requirements.txt`.
  - Added `ktir-lsp-server`.
  - Moved interfaces from `KtdpAttrs.td` to `KtdpAttrInterfaces.td` to resolve downstream TableGen scoping issues.
  - Changed `ktdp_nanobind` module to canonical `ktdp` name to allow MLIR's Python extensions to actually load it when visiting ops.
  - Changed Python package build to allow repackaging the bindings with superprojects (sources also placed in the install tree).
  - Changed `ktir-opt` to also register extensions.


Prasanth-Chatarasi (dismissed):
Looks good to me.

fabianlim:
@Prasanth-Chatarasi @KFAFSP thanks for pinging me on this, let me do a quick local test with the https://github.com/torch-spyre/ktir-cpu/ and https://github.com/torch-spyre/triton that consume this repo

.github/workflows/ci.yml
```diff
@@ -143,7 +143,7 @@ jobs:
       # it via CMAKE_ARGS so both mlir_wheel and custom LLVM paths are uniform.
 
       - name: Build and install Python wheel
-        run: CMAKE_ARGS="-DMLIR_DIR=${{ env.MLIR_DIR }}" uv sync --extra test
+        run: CMAKE_ARGS="-DMLIR_DIR=${{ env.MLIR_DIR }}" uv sync --extra test --inexact
```
fabianlim, line 146: can you clarify why `--inexact` is needed here? If its not obvious then maybe we need some documentation or comment in the ci.yml

fabianlim (dismissed):
I checked it and left one comment, I also setup test PRs for the consumers. Understand that the purpose is to move to the structure of https://github.com/fabianlim/dataflow-scheduler, so I dont have too many comments / objections

I will request also @kiszk to take a look at this

Test PRS:
- https://github.com/torch-spyre/ktir-cpu/pull/145
- https://github.com/torch-spyre/triton/pull/44

.github/workflows/ci.yml
KFAFSP: Of course! In fact, it is quite important that you ascertain whether this is right for you, since it relates to how MLIR is (not being) shipped (anymore).

`uv sync` and `uv run` differ in their behavior, such that `uv sync` without the `--inexact` flag will not only install missing packages in the venv, it will also **remove** all the ones not appearing in the lock file. Since the `mlir` wheel, when it is being used, does not appear in the lock file, this step deletes it. Not only does this make local testing of the pipeline really annoying, it also deletes all the dynamic shared libraries that the build artifact links against.

This wasn't a problem before, since the LLVM and MLIR DSOs were copied into the output directory. However, that couldn't remain as-is, since when building against STATIC or SHARED MLIR (not DYLIB), they don't exist, and also the superproject intends to ship / be handed its own MLIR.

Initially, I added an `IMPORTED_RUNTIME_ARTIFACTS` install target ([see this commit](https://github.com/KFAFSP/ktir-mlir-frontend/commit/af72197ade9295e880a741d1e29382c24ddb606b)), but that didn't seem sound to me either. It is of great importance that all users use the same MLIR library - so having CMake set an RPATH to the install made the most sense.

I don't know how downstream users are consuming this project. It may very well be that they don't otherwise include MLIR and therefore expect the bindings to ship these libraries. For example, to get a self-contained wheel without referencing the MLIR wheel. If so, we should add the install instructions given in my referenced commit above again, but gated behind that specific build configuration. Otherwise, we risk injecting some foreign MLIR instance into the consumer's environment that may already have one.

I'll gladly update the PR based on your requirements in this matter!

KFAFSP:
As a side-note on the C++ downstream users: references to the KTIR libraries should now always be fully-qualified, i.e., `KTIR::KtdpDialect` instead of `KTIRKtdpDialect`. First of all, these names are the same no matter how KTIR is included (`add_subdirectory` vs `find_package`), but CMake will also never interpret them as library names and always target references (failing on configure instead of build when they don't exist).

KFAFSP:
@fabianlim I took the liberty of leaving a few comments on the [torch-spyre/triton](https://github.com/torch-spyre/triton/pull/44) PR on how the build system is supposed to be used. Keep in mind that `EXCLUDE_FROM_ALL` also disable install targets in the subdirectory. You might not want it in that case - but rest assured that documentation is excluded by default.

fabianlim:
@KFAFSP 

> @fabianlim I took the liberty of leaving a few comments on the https://github.com/torch-spyre/triton/pull/44 PR on how the build system is supposed to be used. Keep in mind that EXCLUDE_FROM_ALL also disable install targets in the subdirectory. You might not want it in that case - but rest assured that documentation is excluded by default.

thank you very much i have made the changes [here](https://github.com/torch-spyre/triton/compare/deec060a926c2cd0796b5fb4e774a63dc56a9eae..a0a198a245a5a1c9aab73ebe24e6d3db2591035f)

.github/workflows/ci.yml
KFAFSP: I've implemented my suggestion on how to enable a self-contained wheel build and attached a patch that does it. Maybe this will help you decide.

It introduces a new `KTIR_WHEEL_BUILD` flag that makes it so the library output directory is the `_mlir_libs` folder in the Python package that is generated, sets the RPATH to that directory and installs the runtime components there if needed. I strongly advise against building a wheel with `BUILD_SHARED_LIBS`, though this would then also be possible, provided the `KTIRPythonCAPI` would be updated.

With this patch, the `--inexact` flag is technically not needed anymore, but I'd still argue it should be there such that the build process can be repeated if necessary.

<details>
<summary>Patch for self-contained wheel build</summary>

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 03be8b5..8acf2b1 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,10 +41,12 @@ if(PROJECT_IS_TOP_LEVEL)
   set(LLVM_LIBRARY_OUTPUT_INTDIR ${PROJECT_BINARY_DIR}/lib)
   set(LLVM_RUNTIME_OUTPUT_INTDIR ${PROJECT_BINARY_DIR}/bin)
 
-  # Set up RPATH for installed distribution.
-  include(GNUInstallDirs)
-  set(CMAKE_INSTALL_RPATH_USE_LINK_PATH YES)
-  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+  if(NOT KTIR_WHEEL_BUILD)
+    # Set up RPATH for installed distribution.
+    include(GNUInstallDirs)
+    set(CMAKE_INSTALL_RPATH_USE_LINK_PATH YES)
+    set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+  endif()
 endif()
 
 #
@@ -104,6 +106,20 @@ if(KTIR_ENABLE_PYTHON_BINDINGS)
   endif()
 endif()
 
+# Handle self-contained wheel build.
+if(KTIR_WHEEL_BUILD)
+  set(CMAKE_INSTALL_LIBDIR "${MLIR_BINDINGS_PYTHON_INSTALL_PREFIX}/_mlir_libs")
+  set(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
+  set(CMAKE_PLATFORM_NO_VERSIONED_SONAME ON)
+
+  if(LLVM_LINK_LLVM_DYLIB)
+    install(IMPORTED_RUNTIME_ARTIFACTS LLVM COMPONENT KTIRRuntime)
+  endif()
+  if(MLIR_LINK_MLIR_DYLIB)
+    install(IMPORTED_RUNTIME_ARTIFACTS MLIR COMPONENT KTIRRuntime)
+  endif()
+endif()
+
 # Include KTIR CMake module.
 set(KTIR_INCLUDE_DIRS 
   ${PROJECT_SOURCE_DIR}/include
diff --git a/pyproject.toml b/pyproject.toml
index e83c700..cecd2c0 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,8 +18,8 @@ test = ["pytest", "lit"]
 
 [tool.scikit-build]
 cmake.build-type = "Release"
-cmake.args = ["-DKTIR_ENABLE_PYTHON_BINDINGS=ON", "-GNinja"]
-install.components = ["KTIRPythonModules"]
+cmake.args = ["-DKTIR_ENABLE_PYTHON_BINDINGS=ON", "-DKTIR_WHEEL_BUILD=ON", "-GNinja"]
+install.components = ["KTIRPythonModules", "KTIRRuntime"]
 # mlir_ktdp/ is fully cmake-managed (extensions + MLIR-generated Python).
 # ktdp_tools/ is pure-Python, NOT managed by cmake, so scikit-build-core
 # handles it here.  In editable installs this gives ktdp_tools a live source
```
</details>

cmake/AddKTIR.cmake
```diff
@@ -0,0 +1,138 @@
...
+  cmake_parse_arguments(ARG "" "EXPORT_NAME" "" ${ARGN})
+
+  if(NOT KTIR_BUILD_TOOLS)
+    # Tools are always included, but not necessarily built.
+    set(EXCLUDE_FROM_ALL ON)
+  endif()
+
+  add_llvm_executable(${name} ${ARG_UNPARSED_ARGUMENTS})
+
+  if(KTIR_BUILD_TOOLS AND EXPORT_NAME)
```
kiszk: Is it ok to use `EXPORT_NAME` rather than `ARG_EXPORT_NAME`?
Lines 64, 65, and other lines refer to `ARG_EXPORT_NAME`

tools/ktir-opt/CMakeLists.txt
```diff
@@ -1,14 +1,29 @@
...
+  MLIRRegisterAllExtensions
 )
-add_llvm_executable(ktir-opt ktir-opt.cpp)
-
 llvm_update_compile_flags(ktir-opt)
-target_link_libraries(ktir-opt PRIVATE ${LIBS})
+export_executable_symbols_for_plugins(ktir-opt)
+
+get_target_property(_link ktir-opt LINK_LIBRARIES)
+message(${_link})
```
kiszk: Do we want to leave this line? If it is useful, it is fine with me.

kiszk:
@KFAFSP Thank you. Overall, it is fine. I left two small questions.

cmake/AddKTIR.cmake
KFAFSP: Oh no... Yeah, you're right, thanks! At least, for some weird reason, I made this mistake only in this repo. 

My local build test only tries to use installed libraries, not tools, so that never came up. I should PR the CMake packaging test as well at some point...

tools/ktir-opt/CMakeLists.txt
KFAFSP: That was an oversight from debugging and I don't think it helps anyone. Thanks for catching it!

KFAFSP:
Fixing the DCO required a force-push. Also, @fabianlim please let me know if the self-contained wheel patch that I provided in the discussion above should be included as well or not.

.github/workflows/ci.yml
fabianlim: ok i see. we have two build processes. 
- the default is to use `setup_mlir.py` script to download prebuild llvm artifacts from the github repo. In this case `uv sync` will not touch them. -> `--inexact` is not needed here
- the wheel fallback is a convinience flow say in instances where we want to test with a nightly llvm build, then we use it. -> in this case `--inexact` fix will be needed, so I guess the build will fail in this case. 

> Not only does this make local testing of the pipeline really annoying, it also deletes all the dynamic shared libraries that the build artifact links against.

Im abit confused why this will affect you if you are not using the `--wheel` nightly distro for local builds, or are you using nightly distro for dev?

> It introduces a new KTIR_WHEEL_BUILD flag that makes it so the library output directory is the _mlir_libs folder in the Python package that is generated, sets the RPATH to that directory and installs the runtime components there if needed. I strongly advise against building a wheel with BUILD_SHARED_LIBS, though this would then also be possible, provided the KTIRPythonCAPI would be updated.

My thinking now is that since `--wheel` is intended to be a convinience build im ok to just leave `--inexact` in there. There was never an intention to make this wheel self-contained with mlir binaries.

